PR #2, opened 2020-06-03
Proposal for flytectl

kumare3:


wild-endeavor:
Haven't read yet, but are we doing proposals in Github now?

kumare3:
> Haven't read yet, but are we doing proposals in Github now?
Not really, It made sense in this case, because this can easily become documentation.

brucearctor:
Right now is coded for `flytectl get projects`.  Do we want the plural, singular, or for either (project and/or projects) to be valid?

brucearctor:
Is there a style guide in place here -- wondering why it looks like lines are cut at 200 characters in the README?  (Looks fine, verifying that is desired/purposeful).

kumare3:
I don’t think there is a style guide we have for markdown. But for go we do, we will add the boilerplate to this repo

PR #3, opened 2020-06-03
Initial Commit

EngHabu:


kumare3:
This looks great

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Unshallow
+        run: git fetch --prune --unshallow
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14
+      - name: Run GoReleaser
```
evalsocket, line 19: @EngHabu In that case we will publish binary in each pull request/PR. Ideally this Action should be run when we push a tag

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
+name: goreleaser
+
```
evalsocket, line 2:
```
on:
  push:
    tags:
      - 'v.*.*.*'
```

PR #4, opened 2020-07-13
Update README.md

brucearctor:
Added dev instructions (should this readme be in a different section)?  like /dev/README.md  or other?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,18 @@
 # flytectl
 Flyte CLI
+
+In order to dev, needing to work with a cluster.
+
+A local one can be setup via --> https://lyft.github.io/flyte/administrator/install/getting_started.html#getting-started
+
+
+get the gRPC port:
+```
+FLYTECTL_GRPC_PORT=`kubectl get service -n flyte flyteadmin -o json | jq '.spec.ports[] | select(.name=="grpc").port'`
+```
+
+`kubectl port-forward -n flyte service/flyteadmin 8081:$FLYTECTL_GRPC_PORT`
+
+Update config line in https://github.com/lyft/flytectl/blob/master/config.yaml
+to dns:///localhost:8081 
+


README.md
```diff
@@ -1,2 +1,18 @@
 # flytectl
 Flyte CLI
+
+In order to dev, needing to work with a cluster.
+
+A local one can be setup via --> https://lyft.github.io/flyte/administrator/install/getting_started.html#getting-started
+
+
+get the gRPC port:
```
kumare3, line 9: this is required if Flyte is running in a VPC which is not accessible right?
Otherwise we in sandbox today (locally) we set-up a nodeport service on `locahost:30081`? So maybe we want to clarify this point?

Something like
1. If you have FlyteAdmin hosted behind a load-balancer then - use the DNS 
2. If you are running the standard getting-started sandbox setup locally on docker desktop then - use localhost:30081
3. if you are running sandbox on minikube then run `minikube ip`:30081?
4. if it is running on some k8s cluster not exposed via a LB, then 
your command?

kumare3:
I like the idea of dev readMe. But it can just be the a subsection for now?

README.md
EngHabu: Austin had a problem using the nodeport for GRPC so this was the workaround we did to make it work. Let's actually debug that issue and figure out the proper cause... This can live as a workaround still (in a troubleshooting subsection perhaps?)

README.md
```diff
@@ -1,2 +1,18 @@
 # flytectl
 Flyte CLI
+
+In order to dev, needing to work with a cluster.
```
EngHabu, line 4: Can we move this into a docs/contributing.md file maybe?
I think the main README.md should give a high level view on what this repo is about... maybe installation and building instructions...

README.md
brucearctor: Yes, good to find the underlying issue -- wanted to surface a workaround, in case others experience same

README.md
brucearctor: I'll start a docs/contributing.md with this info.  Seems a good location to also add additional things like building/etc.

PR #9, opened 2020-08-05
Feature/get workflow,Task,Domain,Project

evalsocket:


cmd/get/get.go
```diff
@@ -57,3 +153,101 @@ func toPrintableProjects(projects []*admin.Project) []interface{} {
...
+			Project: p.Project,
+		})
+	}
+
+	return res
+}
+
+func toPrintableGetTask(tasks []*admin.Task) []interface{} {
+	type printableTask struct {
+		Version    string `header:"Version"`
```
kumare3, line 178: linter seems off, I think we should add the same linting rules to this repo as our other repos'

cmd/get/get.go
```diff
@@ -57,3 +153,101 @@ func toPrintableProjects(projects []*admin.Project) []interface{} {
 
 	return res
 }
+
+func toPrintableTask(tasks []*admin.NamedEntityIdentifier) []interface{} {
```
kumare3, line 157: I guess this is what we want to auto-generate, but until then we should add tests?

cmd/get/get.go
```diff
@@ -18,27 +20,121 @@ func CreateGetCommand() *cobra.Command {
...
 		return err
 	}
-
 	logger.Debugf(ctx, "Retrieved %v projects", len(projects.Projects))
 	printer := tableprinter.New(cmdCtx.OutputPipe())
 	printer.Print(toPrintableProjects(projects.Projects))
 	return nil
 }
 
+func getTaskFunc(ctx context.Context, args []string, cmdCtx cmdCore.CommandContext) error {
```
kumare3, line 46: can we split these functions into seprate files, like "get/workflow.go", "get/task.go"

cmd/core/cmd.go
```diff
@@ -1,4 +1,4 @@
-package core
+package cmdCore
```
EngHabu, line 1: typically I would keep the package name matching the directory name any reason not to in this case?

cmd/get/get.go
```diff
@@ -57,3 +153,101 @@ func toPrintableProjects(projects []*admin.Project) []interface{} {
...
+			Name:   w.Name,
+			Domain: w.Domain,
+			Project: w.Project,
+		})
+	}
+
+	return res
+}
+
+func toPrintableListWorkflow(workflows []*admin.Workflow) []interface{} {
```
EngHabu, line 238: Anyway we can change our printing logic to be a bit more generic?
e.g. we can serialize any protobuf into a json object (`map[string]interface{}`) then store somewhere (server side or even just a const here) the list of json paths to use for columns for each object type... then use a library like [this](https://godoc.org/github.com/yalp/jsonpath) to access these fields...

This way you can probably write just `toPrintableObject` and `toPrintableList` functions and have them work on any of the objects we support.

cmd/get/get.go
evalsocket: ok

cmd/get/get.go
evalsocket: ok

cmd/get/get.go
evalsocket: ok

cmd/core/cmd.go
evalsocket: there is a conflict between proto package core and cmd package

cmd/get/get.go
evalsocket: I am working on that. I will create a sperate PR for that....This pr is for feedback as product requirement meaning what data we want to present in table

evalsocket:
@kumare3 @EngHabu I added a approach for printing table please check and let me know if there is any suggestion

brucearctor:
No action on this in last few weeks -- is this something @evalsocket is working on?  Or is this just waiting for @kumare3 to merge?  

I see:  https://github.com/lyft/flyte/milestone/7 which is meant to include 2 of these GET task, GET workflow).  Taking a look now for implementation for GET launchplan and GET execution -- since that is also on the roadmap.  Imagining we'd want uniform/standard/common printing options.

kumare3:
I think @evalsocket  it's working on the jsonpath, if not maybe I can help

evalsocket:
It's done just need testing

kumare3:
> It's done just need testing

Can you push it to the branch we can chip in

cmd/get/project.go
```diff
@@ -0,0 +1,38 @@
...
+			}
+		}
+	}
+	projects, err := cmdCtx.AdminClient().ListProjects(ctx, &admin.ProjectListRequest{})
+	if err != nil {
+		return err
+	}
+	logger.Debugf(ctx, "Retrieved %v projects", len(projects.Projects))
+	adminPrinter := printer.ProjectList{
+		Ctx: cmdCtx,
```
kumare3, line 34: should we pass the commandContext as an input in the PrintMethod? Is there an advantage of making it a class variable. The naming makes it seem like `context` thus counter-intuitive to set it.

cmd/get/workflow.go
```diff
@@ -0,0 +1,54 @@
...
+		adminPrinter := printer.AdminWorkflowsList{
+			Ctx: cmdCtx,
+		}
+		adminPrinter.Print(config.GetConfig().Output, workflows.Workflows)
+		return nil
+	}
+	workflows, err := cmdCtx.AdminClient().ListWorkflowIds(ctx, &admin.NamedEntityIdentifierListRequest{
+		Project: config.GetConfig().Project,
+		Domain:  config.GetConfig().Domain,
+		Limit:   3,
```
kumare3, line 43: 3 seems very small?

pkg/printer/printer.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (p ProjectList) Print(output string, i interface{}) {
+	var projects interface{}
+	byte, _ := json.Marshal(i)
+	_ = json.Unmarshal(byte, &projects)
+	obj := projects.([]interface{})
+	res := make([]interface{}, 0, len(obj))
+	for _, p := range obj {
+		id, _ := jsonpath.Read(p, "$.id")
```
kumare3, line 101: this is great, just that I was thinking that we can take a list of `jsonpaths` as input to print or set it in the context and then just iterate over it printing every one of the set values?

And we could configure the default jsonpath list per entity for example, for project it could be
tableCols = []string{
"$.id",
"$.name",
"$.description",
}

kumare3: we could also include the column header in the same config for example
```
tableCols = map[string]string{
 "Id": "$.id",
"Name": "$.name",
"Description": "$.description",
}

**Comment by kumare3** [comment]
This is awesome, last couple things and then we are set to merge it, maybe a couple unit tests

File: cmd/get/project.go [hunk 9]
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/lyft/flytestdlib/logger"
+)
+var tableStructure = map[string]string{
+	"Id" : "$.id",
+	"Name" : "$.name",
+	"Description" : "$.description",
+}
+
+var entityStructure = map[string]string{
```
kumare3, line 17: where do we use this?

cmd/get/workflow.go
kumare3: you still have 3 here?

pkg/printer/printer.go
```diff
@@ -0,0 +1,151 @@
...
+	case "yaml": // Print protobuf to yaml
+		break
+	default: // Print table
+		printer := tableprinter.New(os.Stdout)
+		printer.Print(i)
+		break
+	}
+}
+
+func (p Printer) PrintBuildNamedEntityIdentifier(output string, i interface{},column map[string]string) error {
```
kumare3, line 26: these functions seem to be repeated? Can we not just use one?

cmd/get/project.go
evalsocket: we needed it but i put it in wrong file

cmd/get/project.go
```diff
@@ -0,0 +1,46 @@
...
+	if len(args) == 1 {
+		projects, err := cmdCtx.AdminClient().ListProjects(ctx, &admin.ProjectListRequest{})
+		if err != nil {
+			return err
+		}
+		logger.Debugf(ctx, "Retrieved %v projects", len(projects.Projects))
+		for _, v := range projects.Projects {
+			if v.Name == args[0] {
+
+				adminPrinter.PrintProject(config.GetConfig().Output, projects.Projects,tableStructure)
```
kumare3, line 30: why do we need `PrintProject` not just a generic `Print`

pkg/printer/printer.go
```diff
@@ -0,0 +1,93 @@
...
+				response = append(response, results)
+				break;
+			case "PrintableProject":
+				results := PrintableProject{}
+				if err := json.Unmarshal(jsonbody, &results); err != nil {
+					return response,err
+				}
+				response = append(response, results)
+				break;
+			case "PrintableNamedEntityIdentifier":
```
kumare3, line 71: I think I am missing something,
I see that you have create a special structure to print a table, but that is the only difference between all "cases"
I want to think of a way of getting rid of the structure for printing a table. That way, all of this code will be once and then we can just start adding all command

@EngHabu you have more context about this library - comments?

evalsocket:
@kumare3 Their is one more alternative of jsonpath. check this https://github.com/jmespath/go-jmespath

kumare3 (approved):
LGTM

kumare3:
I think the linter is not adding spaces? and we need to add some unit tests @evalsocket can you add unit tests, as a separate pass is ok

kumare3 (approved):

PR #15, opened 2020-10-12
Task details in table and using jsonpb to print data

kumare3:
# TL;DR
The default template was wrong and did not refer to the right json-path

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/adminutils/iterator.go
+++ b/adminutils/iterator.go
@@ -3,6 +3,7 @@ package adminutils
 import (
 	"context"
 
+	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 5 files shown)


evalsocket:
@kumare3 it looks good to me

kumare3:
> @kumare3 it looks good to me

@evalsocket I found another bug, we should be using jsonpb to unmarshal, so i will add it to this PR and should be out soon

wild-endeavor (approved):
+1 but didn't look too closely.  let me know if you want an in-depth review.

PR #36, opened 2021-02-18
rename to flyteorg from lyft org, for flytectl

brucearctor:
# TL;DR
Starting to update references to flyteorg

## Type
 - [x] Bug Fix

## Are all requirements met?

 - [x] Code completed

## Complete description
Changed reference links

## Tracking Issue
https://github.com/lyft/flyte/issues/729


cmd/config/config.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	"github.com/lyft/flytestdlib/config"
```
kumare3, line 7: Can we migrate all of them as well
```suggestion
	"github.com/flyteorg/flytestdlib/config"
```

And to make it work without actually changing the upstream you can in `go.mod` add a replace directive

replace (
github.com/flyteorg/flytestdlib => github.com/lyft/flytestdlib/config
)

And do the same for all

kumare3:
@brucearctor I would prefer if we can do this for all upstream dependencies (I think they are flytestdlib, flyteidl) and use the `go.mod` directive

cmd/config/config.go
brucearctor: Reading up on --> https://golang.org/ref/mod#go-mod-file-replace 

I have no experience with replace directive.  But, I had imagined updating ALL of these (without replace directive).  Started with this commit.  But, the others are super easy to follow.  

Do you have a sense of pros/cons, why the replace directive is better (or is it supposedly less typing, since only changing in one place)?  

If we want the same for also flytestdlib, flyteidl as well, easy enough to do.  Are there others?  Or that would cover it?

brucearctor: Ah, I think (if not using go.mod redirect), we need to update each module/repo's self-references.  Then once committed, we should be able to refer out to them (at least is how I am inferring the unit test fail once I tried to make the suggested changes).  It is late, so I could also not be thinking straight ;-)

brucearctor:
Interesting... I anticipate this will pass after some of those other updates are merged to flyteidl, flytestdlib (...again, easily not thinking straight, too -- and the tests certainly must actually pass before we merge).

EngHabu:
Thank you for doing this! If you are going to update all the other repos (and I can't tell you how much I appreciate your effort doing that), then I would do them in this order and avoid the replace directive alltogether:
flytestdlib
flyteidl
flyteconsole,datacatalog,flytectl,
flyteplugins
flytepropeller
flyteadmin

brucearctor:
@EngHabu -- ya, this seems a necessary task that about anyone could do.  So, might as well be me.  (I've got the Ctl stuff down, but still wrapping head around the rest of flyte -- so a way to be actually contributing ;-) )

I think steps (based on your ordering):
* update flytestdlib in that repo and self references to.
* update flyteidl in its own repo (self references) and to flytestdlb
* update in flytestdlib repo references to flyteidl
* update in flyteconsole self references, and to flytestdlib, flyteidl
* etc

We need the module names updated.  That looks like it is a problem across repos for the test/building to pass (until the references within each repo are resolved), so I think each repo needs to update self-references, before the other repos can refer to them... ?  

It seems could just start doing things (actually any order shouldn't matter, and getting any references updated gets us closer).  Ex: this PR. Do we need a larger roadmap, issue, subissues to track things ahead of starting/merging any work?

EngHabu:
Hey @brucearctor, awesome! we do not have circular dependencies between repos. That's why I stated the order I did... if you follow that, you won't need to go back to a repo you already updated as you walk down that list...
When you update flytestdlib, you won't find any references in that repo to other flyte repos... if that makes sense?

brucearctor:
Yes, it does make sense -- though, I also don't care about addressing in stages.  Easy enough to proceed that way.

There is something interesting around: https://github.com/flyteorg/flytestdlib/pull/70/checks?check_run_id=1919791781

And more generally the lyft/boilerplate stuff.  I will try to treat that as separate (but also likely good idea to update)

kumare3:
@brucearctor we can now easily migrate flytectl, as all other repos have been migrated to flyteorg. Whenever you get a chance, please complete this one as well?

brucearctor:
Sure thing.  I'll take care of it this weekend, if not sooner.

brucearctor:
@kumare3  -- closing this PR.  I have another I'm working on.  

So far so good, but I know little about Makefiles and associated linting (which is the error barfing at me), so more to dig into there!

PR #39, opened 2021-02-25
Create executions in flytectl

pmahindrakar-oss:
# TL;DR
Create the executions for given workflow/task in a project and domain.

There are three steps in generating an execution.

- Generate the execution spec file using the get command.
- Update the inputs for the execution if needed.
- Run the execution by passing in the generated yaml file.

The spec file should be generated first and then run the execution using the spec file

`flytectl get tasks -d development -p flytectldemo core.advanced.run_merge_sort.merge  --version v2 --execFile execution_spec.yaml`

The generated file would look similar to this. Also if the file already exist it would prompt you if you need to overwrite.

```
	 iamRoleURN: ''
	 inputs:
	   sorted_list1:
	   - 0
	   sorted_list2:
	   - 0
	 kubeServiceAcct: ""
	 targetDomain: ""
	 targetProject: ""
	 task: core.advanced.run_merge_sort.merge
	 version: "v2"

```

The generated file can be modified to change the parameter values and can be passed to command for execution

```
genExecSpecFile: true
	 iamRoleURN: 'arn:aws:iam::12345678:role/defaultrole'
	 inputs:
	   sorted_list1:
	   - 2
	   - 4
	   - 6
	   sorted_list2:
	   - 1
	   - 3
	   - 5
	 kubeServiceAcct: ""
	 targetDomain: "development"
	 targetProject: "flytesnacks"
	 task: core.advanced.run_merge_sort.merge
	 version: "v2"
```

Notice the source and target domain/projects can be different.
The root project and domain flags of -p and -d should point to task/launch plans project/domain.

`flytectl create execution --execFile execution_spec.yaml -p flytectldemo -d development --targetProject flytesnacks`

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?
Unit tests are pending

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/395

## Follow-up issue
_NA_


cmd/create/execution.go
```diff
@@ -0,0 +1,123 @@
...
+
+- Generate the execution param yaml file containing in the default input parameters for the launch plan.
+- If the default parameters exist then you can modify the default param values in the generated yaml file.
+- Run the execution passing in the generated yaml file.
+
+The following command would generate core.basic.lp.go_greet.inputs.yaml file by using the launch plan provided in argument.
+This would use the latest version of the launch plan.
+::
+
+ bin/flytectl create execution core.basic.lp.go_greet  -d development  -p flytesnacks -g
```
kumare3, line 33: it should also support passing in a task. This is called as a singletask execution.

cmd/create/execution.go
```diff
@@ -0,0 +1,123 @@
...
+`
+)
+
+//go:generate pflags ExecutionConfig --default-var executionConfig
+
+// ProjectConfig Config hold configuration for project create flags.
+type ExecutionConfig struct {
+	File     string `json:"file" pflag:",file for the execution params.If not specified defaults to <launchplan_name>.inputs.yaml"`
+	Version  string `json:"version" pflag:",version of the launch plan to be executed."`
+	GenParam bool   `json:"genParam" pflag:",bool flag to indicate the generation of params file."`
```
kumare3, line 59:
```suggestion
	GenParam bool   `json:"generate-params-file" pflag:",bool flag to indicate the generation of params file."`
```

cmd/create/execution.go
```diff
@@ -0,0 +1,123 @@
...
+		fmt.Printf("run followup command\nflytectl create execution %v -f %v\n", lpName, fileName)
+		return nil
+	}
+	var yamlData map[string]interface{}
+	if yamlData, err = readParamsFromFile(fileName); err != nil {
+		return err
+	}
+	// Convert to Literal map
+	var inputs *core.LiteralMap
+	if inputs, err = createLiteralMapFromParams(yamlData, lp.Spec.DefaultInputs.Parameters); err != nil {
```
kumare3, line 101: should this be `yamlData["inputs"]` As we just want to take the inputs and not the entire yaml

cmd/create/execution.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func createExecutionCommand(ctx context.Context, args []string, cmdCtx cmdCore.CommandContext) error {
+	if len(args) == 0 {
+		return errors.New("pass in the launch plan to execute")
+	}
+	lpName := args[0]
+	var lp *admin.LaunchPlan
+	var err error
+	if lp, err = getLaunchPlan(ctx, lpName, cmdCtx); err != nil {
```
kumare3, line 73: it should be launchplan or task, should we make that as an argument? for example default could be launchplan, but we could say --single-task then it will use task-id

cmd/create/execution.go
```diff
@@ -0,0 +1,123 @@
...
+
+	inputs:
+	  am: false
+	  day_of_week: ""
+	  number: 0
+
+The generated file can be modified to change the parameter values and can be passed to command for execution
+::
+
+ bin/flytectl create execution core.basic.lp.go_greet -f core.basic.lp.go_greet.inputs.yaml  -d development  -p flytesnacks
```
kumare3, line 47: so one problem - In flyte for an execution there are 2 `project and domain`s. One can be the source project/domain where the lauchplan/task is defined and the other can be the target project/domain where the execution is actually triggered.

So these should be the options
```bash
flytectl create execution --project x --domain y --launchPlan lp --targetProject p --targetDomain d
```
this will create an execution in "p-d" for launchplan in "x-y"

Absence of targetProject/targetDomain, we could use the same project/domain
```bash
flytectl create execution --project x --domain y --launchPlan lp
```
this will create an execution in "x-y" for launchplan in "x-y"

ALso for Task
```bash
flytectl create execution --project x --domain y --task task --targetProject p --targetDomain d
```

cmd/create/execution.go
kumare3: Also can this be a string argument, where the user can specify the path/name of the file to be created?

cmd/create/execution_util.go
```diff
@@ -0,0 +1,122 @@
...
+				Version:      executionConfig.Version,
+			},
+		})
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		var lpList *admin.LaunchPlanList
+		lpList, err = cmdCtx.AdminClient().ListLaunchPlans(ctx, &admin.ResourceListRequest{
+			Limit: 10,
```
kumare3, line 36: Limit:1?

kumare3:
Few comments before you go to far ahead

cmd/create/execution.go
pmahindrakar-oss: flytekit i dont see it supporting launching tasks. Let me know if i am missing something here 
https://github.com/flyteorg/flytekit/blob/c02075d472b5587d199630bcfc7f9937673c6a0e/flytekit/clis/flyte_cli/main.py#L1044
The resourceType returned as part of the previous line
lp_id = _identifier.Identifier.from_python_std(urn)
is ignored when passing it to fetch launchPlan.

The UI is doing it by passing in the roles
eg : by launching from the UI i had to pass this role
arn:aws:iam::173840052742:role/flytefunctionaltestsbatchworker-production-iad
This option is not available in flytekit.
Do we want to accept this as part of the yaml file  for flytectl  or as explicit argument when launching an execution for a task .

cmd/create/execution.go
pmahindrakar-oss: So here was my thought,
generateParamsFile is the boolean flag which indicates whether its a 
1] params file generation mode (write) if its true
    and writes to path specified using the File option.
    If the  File option is not specified then it defaults to writing it to <launchplan_name>.inputs.yaml
or 
2] read mode if its false where the path in the File option is used to read the params from the yaml file.
    If the File option is not specified then similar to read case we will try to read <launchplan_name>.inputs.yaml

Let me know what you think

cmd/create/execution.go
pmahindrakar-oss: The function

`readParamsFromFile`
only returns the params from the yaml file and hence we dont need to do that lookup here.
https://github.com/flyteorg/flytectl/pull/39/files#diff-0515dbf3e3fa779cd1f53566cbdada541292ad8641429d5bcb04ceba4117adcbR103

I will rename yamlData to paramsFromYaml to indicate this clearly.

pmahindrakar-oss: The function

`readParamsFromFile`
only returns the params from the yaml file and hence we dont need to do that lookup here.
https://github.com/flyteorg/flytectl/pull/39/files#diff-0515dbf3e3fa779cd1f53566cbdada541292ad8641429d5bcb04ceba4117adcbR103

I will rename yamlData to paramsFromYaml to indicate this clearly.

cmd/create/execution.go
kumare3: yes, for now we will have to accept a role as an input. If CLI is not doing that then its a miss. Dont rely on it.
So I mentioned earlier that we want to use the inputs.yaml file to have more than inputs - it should be all the meta inputs like role - all these fields and probably more in the future - https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/admin/execution.proto#L216-L244

cmd/create/execution.go
kumare3: that makes sense

cmd/create/execution.go
kumare3: can we change it to return a structure that you want to maintain in flytectl - rather than parsing as pure json.
like 
createExecution {
id:
inputs:
blah
}

cmd/create/execution.go
pmahindrakar-oss: Done

cmd/create/execution_util.go
pmahindrakar-oss: Done. Missed this earlier.

cmd/create/execution.go
pmahindrakar-oss: Done. Added both workflow and task for creating execution

cmd/create/execution.go
pmahindrakar-oss: Done. Added the same structure we accept from command line for creating execution from file.

cmd/create/execution.go
pmahindrakar-oss: Done. its now support tasks and workflow executions both.

cmd/create/execution.go
```diff
@@ -0,0 +1,200 @@
...
+
+Usage
+`
+)
+
+//go:generate pflags ExecutionConfig --default-var executionConfig
+
+// ProjectConfig Config hold configuration for project create flags.
+type ExecutionConfig struct {
+	File            string                 `json:"file,omitempty" pflag:",file for the execution params.If not specified defaults to <<workflow/task>_name>.inputs.yaml"`
```
kumare3, line 91: rename the file

cmd/create/execution.go
```diff
@@ -0,0 +1,200 @@
...
+	GenExecSpecFile bool                   `json:"genExecSpecFile,omitempty" pflag:",bool flag to indicate the generation of execution spec file."`
+	Workflow        string                 `json:"workflow,omitempty" pflag:",workflow name for which execution needs to be launched.Either an execution can be for task or workflow."`
+	SourceDomain    string                 `json:"sourceDomain" pflag:",domain from where the workflow or task needs to be fetched for creating the execution.This is mandatory if file is not specified."`
```
kumare3, line 94: we can drop sourceproject and sourcedomain and just use the global project domain as the source.
Also if target should be optional

cmd/create/execution.go
```diff
@@ -0,0 +1,200 @@
...
+	SourceProject   string                 `json:"sourceProject" pflag:",project from where the workflow or task needs to be fetched for creating the execution.This is mandatory if file is not specified."`
+	TargetDomain    string                 `json:"targetDomain" pflag:",project where execution needs to be created.If not specified sourceDomain would be used."`
+	TargetProject   string                 `json:"targetProject" pflag:",project where execution needs to be created.If not specified sourceProject would be used."`
+	Task            string                 `json:"task,omitempty" pflag:",task name for which execution needs to be launched.Either an execution can be for task or workflow."`
+	KubeServiceAcct string                 `json:"kubeServiceAcct" pflag:",kubernetes service account AuthRole for launching execution."`
+	IamRoleURN      string                 `json:"iamRoleURN" pflag:",iam role URN AuthRole for launching execution."`
```
kumare3, line 100:
```suggestion
	IamRoleARN      string                 `json:"iamRoleURN" pflag:",iam role URN AuthRole for launching execution."`
```

cmd/create/execution.go
```diff
@@ -0,0 +1,200 @@
...
+	Version         string                 `json:"version" pflag:",version of the launch plan or task to be fetched for execution.If not specified it would use the latest version."`
```
kumare3, line 101:
```suggestion
	Version         string                 `json:"version" pflag:",[optional]version of the launch plan or task to be fetched for execution.If not specified it would use the latest version."`
```

cmd/create/execution.go
```diff
@@ -0,0 +1,200 @@
...
+	Inputs          map[string]interface{} `json:"inputs" pflag:",inputs in the form of key value pair to be passed for launching task or workflow.If the value is array then pass it using comma separated values."`
```
kumare3, line 102: drop pflag

cmd/create/execution.go
```diff
@@ -0,0 +1,200 @@
...
+
+::
+
+ bin/flytectl create execution  --sourceDomain development  --sourceProject flytectldemo --targetDomain development --targetProject flytesnacks --task core.advanced.run_merge_sort.merge   --genExecSpecFile
+
+The generated file would look similar to this
+
+.. code-block:: yaml
+
+	 genExecSpecFile: true
```
kumare3, line 38:
```suggestion
```
remove from example

cmd/create/execution.go
```diff
@@ -0,0 +1,200 @@
...
+Usage
+`
+)
+
+//go:generate pflags ExecutionConfig --default-var executionConfig
+
+// ProjectConfig Config hold configuration for project create flags.
+type ExecutionConfig struct {
+	File            string                 `json:"file,omitempty" pflag:",file for the execution params.If not specified defaults to <<workflow/task>_name>.inputs.yaml"`
+	GenExecSpecFile bool                   `json:"genExecSpecFile,omitempty" pflag:",bool flag to indicate the generation of execution spec file."`
```
kumare3, line 92:
```suggestion
	GenExecSpecFile bool                   `json:"-,omitempty" pflag:"genExecSpecFile,bool flag to indicate the generation of execution spec file."`
```
maybe this works?

cmd/create/execution_util.go
```diff
@@ -0,0 +1,263 @@
...
+		fmt.Printf("reading all the configuration from file. Other flags on command are ignored as of now\n")
+		// Overwrite the pointer. this is ok since we read all configuration from file.
+		// check with @kumare3 if this is ok.
+		executionConfig, err = readExecConfigFromFile(executionConfig.File)
+		isConfigFromFile = true
+	}
+	return isConfigFromFile, err
+}
+
+func getConfigAndValidate() (*ExecutionParams, error) {
```
kumare3, line 193:
```suggestion
func getConfigAndValidate() (ExecutionParams, error) {
```

cmd/create/execution.go
kumare3, line 92: just move to `flytectl get task` and `flytectl get launchplan`

cmd/create/execution_util.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, fmt.Errorf("unable to read from %v yaml file", executionConfig.File)
+	}
+	executionConfigRead := ExecutionConfig{}
+	if _err = yaml.Unmarshal(data, &executionConfigRead); _err != nil {
+		return nil, _err
+	}
+	return &executionConfigRead, nil
+}
+
+func askForConfirmation(s string) bool {
```
kumare3, line 163: move to common utlity in flytectl, this is a nice function

cmd/create/serialization_utils.go
```diff
@@ -0,0 +1,28 @@
+package create
+
+import (
+	"github.com/flyteorg/flyteidl/clients/go/coreutils"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+)
+
+func MakeLiteralForVariables(serialize map[string]interface{}, variables map[string]*core.Variable) (map[string]*core.Literal, error) {
```
kumare3, line 8: lets put a todo to move to flyteidl

cmd/create/execution.go
pmahindrakar-oss: Done.

cmd/create/execution_util.go
pmahindrakar-oss: Done. Added commandutils

cmd/create/serialization_utils.go
pmahindrakar-oss: Done

cmd/create/execution.go
pmahindrakar-oss: Done.

pmahindrakar-oss: The file generation doesn't exclude  even after removing the pflag. I had to comment during file generation and then uncomment them later

cmd/create/execution.go
pmahindrakar-oss: Done. The problem here is that we can always serialize the sourceProject and sourceDomain from the get command so that during the create we dont have to specify them again. Right now the exec spec file can only be used if the right root project and domain are configured.

cmd/create/execution.go
pmahindrakar-oss: Done.

cmd/create/execution.go
pmahindrakar-oss: Since the file has serialized version field always , We should disallow passing this as an override to avoid any mismatch in the generated file task/lp version against the one passed in. Let me know what you think

cmd/create/execution.go
pmahindrakar-oss: Done

cmd/create/execution_util.go
pmahindrakar-oss: Done

cmd/get/execution_util.go
```diff
@@ -0,0 +1,105 @@
...
+	Workflow        string                 `json:"workflow,omitempty"`
+	Task            string                 `json:"task,omitempty"`
+	Version         string                 `json:"version"`
+	Inputs          map[string]interface{} `json:"inputs"`
+}
+
+func writeExecConfigToFile(executionConfig ExecutionConfig, fileName string) error {
+	d, err := yaml.Marshal(executionConfig)
+	if err != nil {
+		fmt.Printf("error: %v", err)
```
kumare3, line 33: ?
```suggestion
		return err
```

cmd/get/execution_util.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+func createAndWriteExecConfigForTask(task *admin.Task, fileName string) error {
+	var err error
+	executionConfig := ExecutionConfig{Task: task.Id.Name, Version: task.Id.Version}
+	if executionConfig.Inputs, err = getParamMapForTask(task); err != nil {
+		return err
+	}
+	if err = writeExecConfigToFile(executionConfig, fileName); err != nil {
```
kumare3, line 53:
```suggestion
	return writeExecConfigToFile(executionConfig, fileName)
```

cmd/get/execution_util.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+func createAndWriteExecConfigForWorkflow(wlp *admin.LaunchPlan, fileName string) error {
+	var err error
+	executionConfig := ExecutionConfig{Workflow: wlp.Id.Name, Version: wlp.Id.Version}
+	if executionConfig.Inputs, err = getParamMapForWorkflow(wlp); err != nil {
+		return err
+	}
+	if err = writeExecConfigToFile(executionConfig, fileName); err != nil {
```
kumare3, line 65:
```suggestion
	return writeExecConfigToFile(executionConfig, fileName)
```

cmd/get/execution_util.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+	return nil
+}
+
+func getParamMapForTask(task *admin.Task) (map[string]interface{}, error) {
+	paramMap := make(map[string]interface{})
```
kumare3, line 72:
```suggestion
	paramMap := make(map[string]interface{}, len(task.Closure.CompiledTask.Template.Interface.Inputs.Variables))
```

kumare3, line 72: check for empty Inputs, empty Interface

cmd/get/execution_util.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		if paramMap[k], err = coreutils.ExtractFromLiteral(varTypeValue); err != nil {
+			return nil, err
+		}
+	}
+	return paramMap, nil
+}
+
+func getParamMapForWorkflow(lp *admin.LaunchPlan) (map[string]interface{}, error) {
+	paramMap := make(map[string]interface{})
```
kumare3, line 87:
```suggestion
	paramMap := make(map[string]interface{}, len(lp.Spec.DefaultInputs.Parameters))
```

kumare3 (commented):
Couple nits, then merge?

cmd/get/execution_util.go
pmahindrakar-oss: Done

cmd/get/execution_util.go
pmahindrakar-oss: Done

cmd/get/execution_util.go
pmahindrakar-oss: Done

cmd/get/execution_util.go
pmahindrakar-oss: Done

cmd/get/execution_util.go
pmahindrakar-oss: Done

cmd/get/execution_util.go
pmahindrakar-oss: Done

kumare3 (approved):

PR #49, opened 2021-04-03
Oauth support in flytectl

pmahindrakar-oss:

# TL;DR

Config for querying admin

```
admin:
  clientId: flytectl
  endpoint: dns:///demo.nuclyde.io
  authType: Pkce
logger:
  show-source: true
  level: 6
```

```
flytectl get project
Warn: No metricsProvider set for the workqueue
INFO[0000] Using config file: [config.yaml]             
INFO[0000] Config section [root] updated. No update handler registered.  src="viper.go:318"
INFO[0000] Config section [logger] updated. Firing updated event.  src="viper.go:320"
{"json":{"src":"viper.go:320"},"level":"info","msg":"Config section [admin] updated. Firing updated event.","ts":"2021-04-28T21:10:49+05:30"}
{"json":{"src":"viper.go:318"},"level":"info","msg":"Config section [storage] updated. No update handler registered.","ts":"2021-04-28T21:10:49+05:30"}
{"json":{"src":"viper.go:318"},"level":"info","msg":"Config section [adminutils] updated. No update handler registered.","ts":"2021-04-28T21:10:49+05:30"}
{"json":{"src":"client.go:182"},"level":"info","msg":"Instantiating a token source to authenticate against Admin, ID: flytectl","ts":"2021-04-28T21:10:49+05:30"}
{"json":{"src":"token_cache_with_keyring.go:59"},"level":"debug","msg":"fetched token from cache with expiry at 2021-04-29 03:03:19.535504 +0530 IST","ts":"2021-04-28T21:10:51+05:30"}
{"json":{"src":"auth_flow_orchestrator.go:35"},"level":"debug","msg":"got a response from the refresh grant for old expiry 2021-04-29 03:03:19.535504 +0530 IST with new expiry 2021-04-29 03:03:19.535504 +0530 IST","ts":"2021-04-28T21:10:51+05:30"}
{"json":{"src":"client.go:65"},"level":"info","msg":"Initialized Admin client","ts":"2021-04-28T21:10:51+05:30"}
{"json":{"src":"project.go:87"},"level":"debug","msg":"Retrieved 4 projects","ts":"2021-04-28T21:10:51+05:30"}
 --------------- --------------- ---------------------------------- 
| ID (4)        | NAME          | DESCRIPTION                      |
 --------------- --------------- ---------------------------------- 
| flytectldemo  | flytectldemo  | Some description for the project |
 --------------- --------------- ---------------------------------- 
| flyteexamples | flyteexamples | flyteexamples description        |
 --------------- --------------- ---------------------------------- 
| flytesnacks   | flytesnacks   | flytesnacks description          |
 --------------- --------------- ---------------------------------- 
| flytetester   | flytetester   | flytetester description          |
 --------------- --------------- ---------------------------------- 
4 rows

```

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/945

## Follow-up issue
_NA_


pkg/auth/oauth2_client.go
```diff
@@ -0,0 +1,191 @@
...
+	codeVerifier = PkceCodeVerifier
+	PkceCodeVerifier = ""
+	return codeVerifier
+}
+
+
+// A valid oauth2 client (check the store) that additionally requests an OpenID Connect id token
+var clientConf = goauth.Config{
+	ClientID:     "flytectl",
+	ClientSecret: "foobar",
```
EngHabu, line 89: You shouldn't need a secret for a public client, right?

pkg/auth/oauth2_client.go
```diff
@@ -0,0 +1,191 @@
...
+	RedirectURL:  "http://localhost:3846/callback",
+	Scopes:       []string{"all"},
+	Endpoint: goauth.Endpoint{
```
EngHabu, line 92: We should use the metadata URL to discover these... the reason being, when admin is configured to use Okta as an external authorization server, these endpoints will point to okta instead... the only way for the client to know that is if it always fetches (and caches?) `<adminUrl>/.well-known/oauth-authorization-server`

pkg/auth/oauth2_client.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+type FlyteCtlTokenSource struct {
+	accessToken string
+}
+
+func (ts *FlyteCtlTokenSource) Token() (*goauth.Token, error) {
+	t := &goauth.Token{
+		AccessToken: ts.accessToken,
+		Expiry:      time.Now().Add(1 * time.Minute),
```
EngHabu, line 105: We should read that from the access token response... admin sends the expiry, right? if not I can fix that..

pkg/auth/oauth2_client.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (cr InsecurePerRPCCredentials) RequireTransportSecurity() bool {
+	return false
+}
+
+func StartAuthFlow(ctx context.Context) (grpc.CallOption, error) {
+	// ### oauth2 client ###
+	http.HandleFunc("/callback", CallbackHandler(clientConf)) // the oauth2 callback endpoint
+	port := "3846"
```
EngHabu, line 132: There is another endpoint in admin `<adminUrl>/config/v1/flyte_client` that gives you the client_id, redirect_uri, scopes to use/request: https://github.com/flyteorg/flyteadmin/blob/oauth2/pkg/rpc/config/flyte_client.go#L23-L28
This way, the administrator can dictate what the client should use... this is required in case of Okta where we can not hardcode our own clientIds and each administrator will need to generate new ones on their own...

pkg/auth/oauth2_client.go
```diff
@@ -0,0 +1,191 @@
...
+	TokenChannel = make(chan string)
+	if os.Getenv("FLYTE_AUTH_PORT") != "" {
+		port = os.Getenv("FLYTE_AUTH_PORT")
+	}
+	PkceCodeVerifier = generateCodeVerifier(64)
+	PkceCodeChallenge = generateCodeChallenge(PkceCodeVerifier)
+	urlToOpen := clientConf.AuthCodeURL("some-random-state-foobar") + "&nonce=some-random-nonce&code_challenge=" +
```
EngHabu, line 139: I'm sure you will do that but just to make sure it doesn't slip, make sure this is truly random

pkg/auth/oauth2_client.go
```diff
@@ -0,0 +1,191 @@
...
+		PkceCodeChallenge + "&code_challenge_method=S256"
+	go func() {
+		if err := http.ListenAndServe(":"+port, nil); err != nil {
+			logger.Fatal(ctx, "Couldn't start the callback http server on port %v", port)
+		}
+	}()
+	fmt.Println("Please open your webbrowser at " + urlToOpen)
+	_ = exec.Command("open", urlToOpen).Run()
```
EngHabu, line 147: We should probably use this: https://pkg.go.dev/github.com/pkg/browser to ensure it works on all OSes

pkg/auth/oauth2_client.go
pmahindrakar-oss: Done

pkg/auth/oauth2_client.go
pmahindrakar-oss: Yeah.  i have fixed this by using the tokenChannel to pass in the entire object instead of just the accessToken.
In that way we have all the details of the token and i don't have to create a seperate copy of it.

Have created two channels,
1] error channel  used for communicating errors after the callback handler is called.
2] token channel used for communicating the token

pkg/auth/oauth2_client.go
pmahindrakar-oss: Done.

pkg/auth/oauth2_client.go
pmahindrakar-oss: Yeah its not used. I have removed it.

pkg/auth/oauth2_client.go
pmahindrakar-oss: Have handled it in the latest change

pkg/auth/oauth2_client.go
pmahindrakar-oss: Done

cmd/root.go
```diff
@@ -42,6 +42,8 @@ func newRootCmd() *cobra.Command {
 	// --root.project, this adds a convenience on top to allow --project to be used
 	rootCmd.PersistentFlags().StringVarP(&(config.GetConfig().Project), "project", "p", "", "Specifies the Flyte project.")
 	rootCmd.PersistentFlags().StringVarP(&(config.GetConfig().Domain), "domain", "d", "", "Specifies the Flyte project's domain.")
+	rootCmd.PersistentFlags().BoolVarP(&(config.GetConfig().UseAuth), "useAuth", "u", false, "Used when talking with an oauth enabled admin")
```
kumare3, line 45: should this not be part of the flyteadmin config?

pmahindrakar-oss: This i had added temporarily until the useAuth option was available to be used from the admin.
We need Haytham's bind variable change to use the command line option from admin namespace.

For now i am able to use this from config file and hence removing this temporary workaround
```
admin:
  endpoint: dns:///localhost:8089
  insecure: true
  useAuth: true
logger:
  show-source: true
  level: 5
```

pkg/auth/oauth2_client.go
```diff
@@ -0,0 +1,111 @@
...
+type FlyteCtlInsecureTokenSource struct {
+	flyteCtlToken *oauth2.Token
+}
+
+func (ts *FlyteCtlInsecureTokenSource) Token() (*oauth2.Token, error) {
+	t := ts.flyteCtlToken
+	return t, nil
+}
+
+type InsecurePerRPCCredentials struct {
```
EngHabu, line 95: We should use this instead: https://github.com/flyteorg/flyteidl/blob/master/clients/go/admin/token_source.go#L36

Since it supports custom authorization header... that may be required in some environments where there are other filters that intercept the default `authorization` header.

pmahindrakar-oss: Sure . Will update this .

pmahindrakar-oss: Done

pkg/auth/auth_flow_orchestrator.go
```diff
@@ -0,0 +1,145 @@
+package auth
+
+import (
```
pmahindrakar-oss, line 3: More Unit tests remaining for this package and will be added soon.

cmd/create/execution.go
```diff
@@ -140,9 +143,16 @@ func createExecutionCommand(ctx context.Context, args []string, cmdCtx cmdCore.C
...
-	if _err != nil {
-		return _err
+	var callOptions []grpc.CallOption
+	var exec *admin.ExecutionCreateResponse
+	grpcAPICall := func(_ctx context.Context, _callOptions []grpc.CallOption) error {
+		var err error
+		exec, err = cmdCtx.AdminClient().CreateExecution(_ctx, executionRequest, _callOptions...)
+		return err
+	}
+	err = auth.Do(ctx, cmdCtx.AuthClient(), grpcAPICall, callOptions)
```
EngHabu, line 153: Is there a way to simplify this?
Like [here](https://github.com/flyteorg/flyteidl/pull/155/files#diff-017baee1fb74ab319b7caf1ef76d1321474dd5aec808c34b0a0f4f98b18c6878R179-R203)
You can build the authmetadataclient without initializing that singleton connection, then call FlyteClient() and if that returns then you can setup a PerCallRPCCredentials and set it as a dial option (instead of a call option) ... and therefore, you do not need to surround every single call to admin with this...

pkg/auth/admin_grpc_call_util.go
```diff
@@ -0,0 +1,84 @@
...
+	if tokenCallOption, err = callOptionForToken(ctx, authMetadataClient, cachedOrRefreshedToken); err != nil {
+		return nil
+	}
+	return append(callOptions, tokenCallOption)
+}
+
+func Do(ctx context.Context, authMetadataClient service.AuthMetadataServiceClient, grpcAPICallContext AdminGrpcAPICallContext, callOptions []grpc.CallOption) error {
+	// Fetch from the cache only when usAuth is enabled.
+	useAuth := admin.GetConfig(ctx).UseAuth
+	if useAuth {
```
EngHabu, line 70: I'm getting rid of this... if we can attempt to setup auth before we Dial(), I think that will simplify the experience a ton....

pkg/auth/auth_flow_orchestrator.go
```diff
@@ -0,0 +1,145 @@
...
+
+const (
+	Timeout     = 15 * time.Second
+	RefreshTime = 5 * time.Minute
+)
+
+type TokenOrchestrator struct {
+}
+
+// This is a copy of oauth2.internal.tokenJSON as its not accesible outside.
```
EngHabu, line 26: nit:
```suggestion
// This is a copy of oauth2.internal.tokenJSON as it's not accesible outside.
```

pkg/auth/auth_flow_orchestrator.go
```diff
@@ -0,0 +1,145 @@
...
+func (e *tokenJSON) expiry() (t time.Time) {
+	if v := e.ExpiresIn; v != 0 {
+		return time.Now().Add(time.Duration(v) * time.Second)
+	}
+	return
+}
+
+func (f TokenOrchestrator) RefreshTheToken(ctx context.Context, clientConf *oauth2.Config, token *oauth2.Token) *oauth2.Token {
+	// ClientSecret is empty here. Basic auth is only needed to refresh the token.
+	client := newBasicClient(clientConf.ClientID, clientConf.ClientSecret)
```
EngHabu, line 46: This is concerning.... public clients should not need or attempt to keep secrets.... if that means we can't refresh tokens, then so be it... we can make the default access token expire in 2 hours (or so) to give people a slightly improved experience...

EngHabu: Or is the idea that this flow will be used in CI/CD system instead?

pkg/auth/auth_flow_orchestrator.go
```diff
@@ -0,0 +1,145 @@
...
+	payload := url.Values{
+		"grant_type":    {"refresh_token"},
+		"refresh_token": {token.RefreshToken},
+		"scope":         {"all", "offline"},
+	}
+	_, body, err := client.Post(clientConf.Endpoint.TokenURL, payload)
```
EngHabu, line 52: can't we use oauth2.Config to Exchange and refresh tokens?

pkg/auth/handler_app_call_back.go
```diff
@@ -0,0 +1,52 @@
...
+		var opts []oauth2.AuthCodeOption
+		opts = append(opts, oauth2.SetAuthURLParam("code_verifier", codeVerifier))
+
+		token, err := c.Exchange(context.Background(), req.URL.Query().Get("code"), opts...)
+		if err != nil {
+			errorChannel <- fmt.Errorf("error while exchanging auth code due to %v", err)
+			_, _ = rw.Write([]byte(fmt.Sprintf(`<p>Couldn't get access token due to error: %s</p>`, err.Error())))
+			return
+		}
+		_, _ = rw.Write([]byte(`<p>Cool! Your authentication was successful and you can close the window.<p>`))
```
EngHabu, line 49: If we want to go fancy, I would say we should do something like this: https://blakewilliams.me/posts/handling-macos-url-schemes-with-go
(which seems to be supported in OSX and windows... not sure about general Linux support) to silently go back to the app once the user is authenticated/authorized...

but maybe that's too fancy?

EngHabu (commented):
Awesome! I have some questions about the token exchange stuff and potentially simplifying the code (that surrounds each api call)... otherwise looks ready (minus those unit tests you are already adding)... thank you!

pkg/auth/auth_flow_orchestrator.go
pmahindrakar-oss: Actually internally oauth libraries do the same when they refresh the token by sending an empty client secret. I have removed this logic and instead using 

```
/ Token returns the current token if it's still valid, else will
// refresh the current token (using r.Context for HTTP client
// information) and return the new one.
func (s *reuseTokenSource) Token() (*Token, error) {

```
This automatically refreshes the token if the access token is expired.

pkg/auth/auth_flow_orchestrator.go
pmahindrakar-oss: Removed this logic now and uses
```func (s *reuseTokenSource) Token() (*Token, error) {
```

pkg/auth/admin_grpc_call_util.go
pmahindrakar-oss: Done

cmd/create/execution.go
pmahindrakar-oss: Done. Created new MR for this https://github.com/flyteorg/flyteidl/pull/156/

pmahindrakar-oss: Will be updating this MR with just changes to flyteidl and couple of modules to be used. Rest there are no changes in flytectl side with this new movement

EngHabu (approved):

kumare3 (approved):

kumare3 (approved):

pkg/auth/handler_app_call_back.go
pmahindrakar-oss: Looks cool . Let me add try this in a followup

PR #54, opened 2021-04-16
Added update for named entity

pmahindrakar-oss:

# TL;DR

Following command updates the description on the workflow.

```
 flytectl update workflow -p flytectldemo -d development core.advanced.run_merge_sort.merge_sort --description "Mergesort workflow example"
```

Archiving workflow named entity would cause this to disapper from flyteconsole UI.

```
 flytectl update workflow -p flytectldemo -d development  core.advanced.run_merge_sort.merge_sort --archive
```

Activating workflow named entity would unarchive it.
```
 flytectl update workflow -p flytectldemo -d development  core.advanced.run_merge_sort.merge_sort --activate
```

Following command updates the description on the task.

```
 flytectl update  task -d development -p flytectldemo core.advanced.run_merge_sort.merge --description "Merge sort example"
```

Archiving task named entity would is not supported and would throw an error.
```
 flytectl update  task -d development -p flytectldemo core.advanced.run_merge_sort.merge --archive
```

Activating workflow named entity would be a noop as archiving is not possible.

```
 flytectl update  task -d development -p flytectldemo core.advanced.run_merge_sort.merge --activate
```

Following command updates the description on the launchplan.

```
 flytectl update launchplan -p flytectldemo -d development  core.advanced.run_merge_sort.merge_sort --description "Mergesort example"
```

Archiving launchplan named entity is not supported and would throw an error.

```
 flytectl update launchplan -p flytectldemo -d development  core.advanced.run_merge_sort.merge_sort --archive
```

Activating launchplan named entity would be a noop.
```
 flytectl update launchplan -p flytectldemo -d development  core.advanced.run_merge_sort.merge_sort --activate
```

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/919


clierrors/errors.go
```diff
@@ -0,0 +1,17 @@
+package clierrors
+
+var (
+	ErrInvalidUpdate = "Invalid state passed. Specify either activate or archive\n"
```
katrogan, line 4: nit: maybe give this a more specific name since it references state exclusively?

cmd/update/interfaces/updater.go
```diff
@@ -0,0 +1,14 @@
...
+	"context"
+
+	cmdCore "github.com/flyteorg/flytectl/cmd/core"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+)
+
+//go:generate mockery -name=Updater -case=underscore
+
+type Updater interface {
+	UpdateNamedEntity(ctx context.Context, name string, project string, domain string, rsType core.ResourceType, cmdCtx cmdCore.CommandContext) error
```
katrogan, line 13: not sure if you already know this, but fyi for brevity you can also do `UpdateNamedEntity(ctx context.Context, name, project, domain string, rsType core.ResourceType, cmdCtx cmdCore.CommandContext) error`

cmd/update/project.go
```diff
@@ -65,23 +67,18 @@ Usage
...
-	errInvalidUpdate   = "Invalid state passed. Specify either activate or archive\n"
-	errFailedUpdate    = "Project %v failed to get updated to %v state due to %v\n"
-)
+var projectConfig = &ProjectConfig{}
 
 func updateProjectsFunc(ctx context.Context, args []string, cmdCtx cmdCore.CommandContext) error {
 	id := config.GetConfig().Project
 	if id == "" {
-		fmt.Printf(errProjectNotFound, id)
+		fmt.Printf(clierrors.ErrProjectNotPassed)
```
katrogan, line 75: why not return an error with the same message?

katrogan (commented):
have you considered raising an error for no-op commands (like activating a task, etc). I wonder if silently doing nothing is preferable as opposed to being more explicit

pmahindrakar-oss:
Hi @katrogan  i did consider raising an error for noop cases , but wont it be preferable to have this coming from admin at this line 
```
if request.Metadata.State != admin.NamedEntityState_NAMED_ENTITY_ACTIVE &&
		request.ResourceType != core.ResourceType_WORKFLOW {
		return errors.NewFlyteAdminErrorf(codes.InvalidArgument,
			"Only workflow name entities can have their state updated")
	}
```

https://github.com/flyteorg/flyteadmin/blob/2d81c1eec24cffb43346b56fc0017fd29db33a38/pkg/manager/impl/validation/named_entity_validator.go#L33

Since we already do it for archive state hence i thought why not have it for other state update checks aswell in admin to make it similar .
Let me know what you think

cmd/update/project.go
pmahindrakar-oss: Missed it . Yes we should be returning this error.
Fixed it.

clierrors/errors.go
pmahindrakar-oss: Done.

cmd/update/interfaces/updater.go
pmahindrakar-oss: Sure . Done. Thanks for pointing that out

kumare3:
@pmahindrakar-oss i dont know if I really prefer `activate` as the opposite of archive - this could be confusing, as activate for launch plan has an alternate meaning.
@katrogan wdyt?

kumare3:
besides the one thing, I am +1 to merge it

pmahindrakar-oss:
I will reword  in documentation to make it clear.
The following statement might have caused this 
```
Activating workflow named entity would unarchive it.

 flytectl update workflow -p flytectldemo -d development  core.advanced.run_merge_sort.merge_sort --activate

```
Will keep those separate

katrogan:
@kumare3 not sure what a better name would be. in the proto def we also call it activate but agreed that word is potentially overloaded

katrogan:
> Hi @katrogan i did consider raising an error for noop cases , but wont it be preferable to have this coming from admin at this line
> 
> ```
> if request.Metadata.State != admin.NamedEntityState_NAMED_ENTITY_ACTIVE &&
> 		request.ResourceType != core.ResourceType_WORKFLOW {
> 		return errors.NewFlyteAdminErrorf(codes.InvalidArgument,
> 			"Only workflow name entities can have their state updated")
> 	}
> ```
> 
> https://github.com/flyteorg/flyteadmin/blob/2d81c1eec24cffb43346b56fc0017fd29db33a38/pkg/manager/impl/validation/named_entity_validator.go#L33
> 
> Since we already do it for archive state hence i thought why not have it for other state update checks aswell in admin to make it similar .
> Let me know what you think

@pmahindrakar-oss I see, so even though it's technically a no-op we'll raise the error that admin returns?

pmahindrakar-oss:
Yes thats right @katrogan . Instead of adding a new validation it would be ok to jus re-throw the new error and i have currently worded it to be noop in the doc since there is no feedback given but once we have the error thrown from admin then we can change this to say that this is an error case and the user should be able to see similar error like archiving launchplan.

kumare3:
@katrogan / @pmahindrakar-oss - So eventually we probably want archive for all entities right? - LaunchPlan and Tasks. So there are 3 ways of solving this
1. Do not expose this functionality in flytectl
2. Expose the functionality but raise an error on client side
3. Expose the functionality and also make an admin call, but let admin return an error

IMO, 3 is the most long term clean solution. But we need to determine if there is a case in which entity archival does not make sense. Can you guys resolve this?

katrogan:
#3 sounds fine

katrogan (approved):

pmahindrakar-oss:
Thanks for the review. Merging it now and we will take of no.3 using the following issue https://github.com/flyteorg/flyte/issues/943

PR #61, opened 2021-05-04
Refactored the fetcher interface to exclude cmdCtx

pmahindrakar-oss:

# TL;DR
Removed the dependency on cmdCtx on the fetcher interface and moved this to commandutlls package along with its mocks and impl

The Fetcher interface allows the command to fetch data that is required for it.
Eg : create execution depends on the fetcher to fetch the launchplan
The get commands themselves use this for fetching the data.

The Fetcher is composed of the adminServiceClient and hence it removes this explicit dependencies in all functions that need to call Admin API's for fetching the data.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_


pkg/commandutils/interfaces/fetcher_interface.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
+)
+
+//go:generate mockery -all -case=underscore
+
+// Fetcher Interface for exposing the fetch capabilities from the admin and also allow this to be injectable into other
+// modules. eg : create execution which requires to fetch launchplan details to construct the execution spec.
+type Fetcher interface {
```
kumare3, line 13: This is very counter intuitive I feel. We have to implement all methods of Admin in here, then why need it. We already have an interface for AdminService?

pmahindrakar-oss: This acts an Extension interface for getting data using AdminServiceClient for flytectl.

Advantages of using this pattern.
1] Simple interface and directly maps to use cases of the commands (eg : all version, latest version or given version)
    This can also be extended to be driven from the config . 
    eg : GetLPUsingCommandConfig which helps to translate the command level config and call the admin api's appropriately
2] Easier testing . The consuming command function become more testable with clear higher level interface instead of relying on admin service mocks. Also reduces testing code and makes sure unit test dont span multiple files as much as possible.
3] Dependency can be used by multiple commands for using the fetch capabilities and they need not redefine or have a dependency on another commands utility functions.

cmd/core/cmd_ctx.go
```diff
@@ -3,17 +3,20 @@ package cmdcore
...
 
+	"github.com/flyteorg/flytectl/pkg/ext"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/service"
 )
 
 type CommandContext struct {
-	adminClient service.AdminServiceClient
-	in          io.Reader
-	out         io.Writer
+	adminClient           service.AdminServiceClient
```
kumare3, line 11: remove this one?

kumare3 (approved):

cmd/core/cmd_ctx.go
pmahindrakar-oss: Will take this in my matchable entity pr . As changing this now will discard your approval

PR #66, opened 2021-05-12
Added support for execution queue attribute

pmahindrakar-oss:

# TL;DR
This PR specifically addresses supporting execution queue attributes through flytectl. 

- **GET**

Retrieves execution  queue attributes for given project,domain combination or additionally with workflow name.

Here the command get execution queue attributes for  project flytectldemo and development domain.

` flytectl get execution-queue-attribute -p flytectldemo -d development `

eg : O/P
```
{"project":"flytectldemo","domain":"development","tags":["buzz","lightyear","foo","bar"]}
```

Writing the execution queue attribute to a file. If there are no execution queue attributes , command would return an error
Here the command gets exectuin queue attributes and writes the config file to era.yaml
eg:  content of tra.yaml

` flytectl get execution-queue-attribute --attrFile era.yaml`

```
.. code-block:: yaml

	domain: development
	project: flytectldemo
	tags:
	  - foo
	  - buzz
	  - lightyear
```

-  **UPDATE**
Updates execution queue attributes for given project and domain combination or additionally with workflow name.

Updating the execution queue attribute is only available  from config file
Here the command updates takes the input for execution queue attributes from the config file era.yaml

eg:  content of era.yaml
`flytectl update execution-queue-attribute -attrFile era.yaml`

- **DELETE**
Deletes execution queue attributes for given project and domain combination or additionally with workflow name.
Deletes execution queue attribute for project and domain
Here the command delete execution queue attributes for  project flytectldemo and development domain.

` flytectl delete execution-queue-attribute -p flytectldemo -d development `

Deleting execution queue attribute using config file
Here the command deletes execution queue attributes from the config file era.yaml
eg:  content of era.yaml which will use the project domain and workflow name for deleting the resource

 `flytectl delete execution-queue-attribute --attrFile era.yaml`

```
.. code-block:: yaml

	domain: development
	project: flytectldemo
	tags:
	  - foo
	  - buzz
	  - lightyear
```

Deleting execution queue attribute for a workflow
Here the command deletes execution queue attributes for a workflow

`flytectl delete execution-attribute -p flytectldemo -d development core.control_flow.run_merge_sort.merge_sort`

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [X] Code documentation added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/992

## Follow-up issue
_NA_


cmd/config/subcommand/executionqueueattribute/file_config.go
```diff
@@ -0,0 +1,47 @@
...
+	return &admin.MatchingAttributes{
+		Target: &admin.MatchingAttributes_ExecutionQueueAttributes{
+			ExecutionQueueAttributes: a.ExecutionQueueAttributes,
+		},
+	}
+}
+
+// UnDecorate to uncover ExecutionQueueAttributes.
+func (a *AttrFileConfig) UnDecorate(matchingAttribute *admin.MatchingAttributes) {
+	if matchingAttribute == nil {
```
katrogan, line 28: when does this happen?

cmd/delete/matchable_execution_queue_attribute.go
```diff
@@ -0,0 +1,81 @@
...
+Deletes execution queue attributes for given project and domain combination or additionally with workflow name.
+
+Deletes execution queue attribute for project and domain
+Here the command delete execution queue attributes for project flytectldemo and development domain.
+::
+
+ flytectl delete execution-queue-attribute -p flytectldemo -d development 
+
+
+Deleting execution queue attribute using config file which was used for creating it.
```
katrogan, line 25:
```suggestion
Deletes execution queue attribute using config file which was used for creating it.
```

cmd/delete/matchable_execution_queue_attribute.go
```diff
@@ -0,0 +1,81 @@
...
+Here the command deletes execution queue attributes from the config file cra.yaml
```
katrogan, line 26: nit: why is this called cra? (copy paste from cluster resources attributes?)

cmd/delete/matchable_execution_queue_attribute.go
```diff
@@ -0,0 +1,81 @@
...
+	domain: development
+	project: flytectldemo
+	tags:
+		- foo
+		- bar
+		- buzz
+		- lightyear
+
+Deleting execution queue attribute for a workflow
+Here the command execution queue attributes for a workflow
```
katrogan, line 45: this looks like it's missing a second half?

cmd/get/matchable_execution_queue_attribute.go
```diff
@@ -0,0 +1,80 @@
...
+	executionQueueAttributesLong  = `
+Retrieves execution queue attributes for given project and domain combination or additionally with workflow name.
+
+Retrieves execution queue attribute for project and domain
+Here the command get execution queue attributes for  project flytectldemo and development domain.
+::
+
+ flytectl get execution-queue-attribute -p flytectldemo -d development 
+
+eg : O/P
```
katrogan, line 24: what is O/P?

cmd/get/matchable_execution_queue_attribute.go
```diff
@@ -0,0 +1,80 @@
...
+const (
+	executionQueueAttributesShort = "Gets matchable resources of execution queue attributes"
+	executionQueueAttributesLong  = `
+Retrieves execution queue attributes for given project and domain combination or additionally with workflow name.
+
+Retrieves execution queue attribute for project and domain
+Here the command get execution queue attributes for  project flytectldemo and development domain.
+::
+
+ flytectl get execution-queue-attribute -p flytectldemo -d development
```
katrogan, line 22: can you add an example with a workflow name too?

docs/source/gen/flytectl_delete_execution-queue-attribute.rst
```diff
@@ -0,0 +1,111 @@
...
+
+Deletes execution queue attribute for project and domain
+Here the command delete execution queue attributes for project flytectldemo and development domain.
+::
+
+ flytectl delete execution-queue-attribute -p flytectldemo -d development 
+
+
+Deleting execution queue attribute using config file which was used for creating it.
+Here the command deletes execution queue attributes from the config file cra.yaml
```
katrogan, line 23:
```suggestion
Here the command deletes execution queue attributes from the config file era.yaml
```

docs/source/gen/flytectl_delete_execution-queue-attribute.rst
```diff
@@ -0,0 +1,111 @@
...
+eg:  content of cra.yaml which will use the project domain and workflow name for deleting the resource
```
katrogan, line 24: Can we note that the tags are optional in the file (since they're unread)?

katrogan (approved):

cmd/config/subcommand/executionqueueattribute/file_config.go
pmahindrakar-oss: This will happen in case where admin doesn't return an error of empty matching attribute and we try to unwrap the response to get the executionQueueAttribute which wont exist.
This is check is safeguard against it.

This is what is passed to undecorate 
`workflowAttr.GetAttributes().GetMatchingAttributes()`
Where workflowAttr is response from the admin 

Ideally it would be preferable to do null checks for both workflowAttr.GetAttributes() and then workflowAttr.GetAttributes().GetMatchingAttributes()

Since the idl can return nil for both
```

func (m *WorkflowAttributesGetResponse) GetAttributes() *WorkflowAttributes {
	if m != nil {
		return m.Attributes
	}
	return nil
}

...

func (m *WorkflowAttributes) GetMatchingAttributes() *MatchingAttributes {
	if m != nil {
		return m.MatchingAttributes
	}
	return nil
}```

cmd/delete/matchable_execution_queue_attribute.go
pmahindrakar-oss: Done.

cmd/delete/matchable_execution_queue_attribute.go
pmahindrakar-oss: Thanks done

cmd/delete/matchable_execution_queue_attribute.go
pmahindrakar-oss: yup . thanks for pointing it out. Fixed it. 
Would need a code generator for comments for such cases.

cmd/get/matchable_execution_queue_attribute.go
pmahindrakar-oss: Done. Added for this and other attributes aswell.

cmd/get/matchable_execution_queue_attribute.go
pmahindrakar-oss: Reworded it
`eg : output from the command`

docs/source/gen/flytectl_delete_execution-queue-attribute.rst
pmahindrakar-oss: Done. Thanks

docs/source/gen/flytectl_delete_execution-queue-attribute.rst
pmahindrakar-oss: Done.
`Tags are optional in the file as they are unread during the delete command but can be kept as the same file can be used for get, update or delete `
Added to other commands aswell

cmd/config/subcommand/executionqueueattribute/file_config.go
katrogan: oh interesting, so there are cases where we don't return not found but just an empty response? that seems wrong

katrogan (approved):

cmd/config/subcommand/executionqueueattribute/file_config.go
pmahindrakar-oss: We might be avoiding this in code path, but the idl supports returning nil hence added a guard .
I haven't seen it during my testing since if there are no attributes its always an error response and if thats the case then we don't call Undecorate.

If the resource is not found we do return an from the ORM layer
https://github.com/flyteorg/flyteadmin/blob/db32bb3810e7490e04ff693eadd26414aa294482/pkg/repositories/gormimpl/resource_repo.go#L114

And which is propogated back to client 
https://github.com/flyteorg/flyteadmin/blob/db32bb3810e7490e04ff693eadd26414aa294482/pkg/manager/impl/resources/resource_manager.go#L220

Hence i don't see how we can end up in this state no error and empty response.

katrogan: great, just wanted to confirm you hadn't actually seen this behavior! protecting against it since it's technically allowed by the response structure sounds fine

PR #73, opened 2021-05-27
Added transparent Get call when fetching latest version of flyte objects [task | launchplan | workflow]

pmahindrakar-oss:

# TL;DR

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1053

## Follow-up issue
_NA_


pmahindrakar-oss:
Before making changes to admin to make it consistent across all flyte objects would be good to know if any clients like flytekit rely on the full object from the List call . cc : @katrogan

katrogan:
@pmahindrakar-oss do we need to make things consistent? for example, executions require a separate call to fetch their data (inputs & outputs) if a launch plan closure is lightweight enough to return inline i think we should do it and only depart from that model when closure data is large

kumare3:
@katrogan / @pmahindrakar-oss I do not follow the discussion? why do we need to make any changes to admin?

katrogan:
@kumare3/ @pmahindrakar-oss  I'm not sure either, I think we should leave admin as is and just have flytectl fetch the individual workflow in a Get call

pmahindrakar-oss:
@kumare3 @katrogan I have kept this flytectl PR to do a transparent get only for workflow --latest version.

katrogan (approved):

PR #75, opened 2021-05-28
Visualize graphs on command line

kumare3:
Signed-off-by: Ketan Umare <ketan.umare@gmail.com>

# TL;DR
Uses graphviz to visualize flyte workflows on commandline. Example of branch node
nested-conditionals
coin-toss
sub-workflows

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Uses graphviz to visualize flyte workflows on commandline


jsonporter:
Graph looks good (once tests pass)

pkg/visualize/graphviz.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func constructTaskNode(name string, graph *cgraph.Graph, n *core.Node, t *core.CompiledTask) (*cgraph.Node, error) {
+	gn, err := graph.CreateNode(name)
+	if err != nil {
+		return nil, err
+	}
+	if n.Metadata != nil && n.Metadata.Name != "" {
+		v := strings.LastIndexAny(n.Metadata.Name, ".")
+		gn.SetLabel(fmt.Sprintf("%s [%s]", n.Metadata.Name[v+1:], t.Template.Type))
```
katrogan, line 79: nit: in the examples it looks like the task type always goes outside the bounding box for the task node - is this fixable here?

kumare3: hmm, fixable probably by using a better way of rendering

katrogan: 💯 sounds good!

evalsocket:
@EngHabu This pr need CGO enable in go that's why it's failing. Goreleaser doesn't support CGO build. I explored some solution around this problem. Community is building CGO enabled build in container, They use different images for different PLATFORM/ARC and in our case it become complicated because we are generating multiple binary and may be we will increase new binaries in future 

second problem is that we use goreleaser for generating `install.sh` that we use in our docker images for installing flytectl, That will break

Third problem is brew, Currently brew package is published by goreleaser or else we need to do this manually 


Need your thoughts on this. Alternative option is to use different library for graph

kumare3:
@evalsocket this is really sad. The reason I used this library is because it brings the c-bindings with it. I.e no need of external dot library. Can we limit some go packages only for some platforms like osx

evalsocket:
not sure.  This lib look strong https://github.com/awalterschulze/gographviz. I discussed with @pmahindrakar-oss and he is exploring the capabilities of this lib.

kumare3:
The lib will only print a dot graph - and users will have to use an online renderer

kumare3:
I guess maybe we have to do this

evalsocket:
ok

evalsocket:
@kumare3 @pmahindrakar-oss checkout influxdb release process. They are using goreleaser for releasing binary with CGO enabled https://github.com/influxdata/influxdb/blob/master/.goreleaser.yml

kumare3:
We can try in this or

pmahindrakar-oss:
@evalsocket let us know how it goes with the CGO changes similar to influxdb which you have suggested.

The changes with other library are also almost done aswell which dumps the dot string which can be used in online or commandline renderer for dot files. I will wait for your results to see if we want to change to complete go library for graphviz or use the existing one.

I used this for eg :

tinyurl.com/h6m7559j

pmahindrakar-oss (approved):

PR #85, opened 2021-06-05
Added Github workflow for generating docs

evalsocket:
# TL;DR
- Added GitHub workflow for generating documentation on every master merge.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
 _How did you fix the bug, make the feature, etc. Link to any design docs etc_

## Tracking Issue
https://github.com/lyft/flyte/issues/1100

## Follow-up issue
_NA_
OR


.github/workflows/generate-docs.yaml
```diff
@@ -0,0 +1,54 @@
...
+          path: ~/.cache/pip
+          key: docs-pip-${{ runner.os }}-${{ hashFiles('doc-requirements.txt') }}-${{ hashFiles('doc-requirements.in') }}
+          restore-keys: docs-pip-
+      - name: Install Dependencies
+        run: |
+          # Install all requirments
+          make doc-requirements.txt
+      - name: Generate documentation
+        run: |
+          make -C docs gendocs
```
evalsocket, line 33: It is failing because the dev tool is not installed. Like `sphinx-build`. Any idea why?

cmd/get/execution_util.go
```diff
@@ -125,4 +125,4 @@ func ParamMapForWorkflow(lp *admin.LaunchPlan) (map[string]interface{}, error) {
 		}
 	}
 	return paramMap, nil
-}
+}
```
pmahindrakar-oss, line 128: Change can be removed from this PR

.github/workflows/generate-docs.yaml
pmahindrakar-oss: Can we bring in the doc requirements from flytesnacks , or better can that doc-requirements be moved to boilerplate and be used in all repos . So that all the repo where we generate docs have these dependencies pulled in and built

cmd/get/execution_util.go
evalsocket: ok

.github/workflows/generate-docs.yaml
```diff
@@ -0,0 +1,54 @@
...
+          python-version: 3.9
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: docs-pip-${{ runner.os }}-${{ hashFiles('doc-requirements.txt') }}-${{ hashFiles('doc-requirements.in') }}
+          restore-keys: docs-pip-
+      - name: Install Dependencies
+        run: |
+          # Install all requirments
+          make doc-requirements.txt
```
EngHabu, line 30: This doesn't actually install the requirements, it updates the file with latest values of requirement.
you will need to run `pip install -r doc-requirements.txt` to install...

.github/workflows/generate-docs.yaml
```diff
@@ -0,0 +1,54 @@
...
+          lfs: true
+      - uses: actions/setup-go@v1
+        with:
+          go-version: '1.16'
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.9
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
```
EngHabu, line 24: Is this the path pip actually uses to cache requirements?

.github/workflows/generate-docs.yaml
evalsocket: Thanks

.github/workflows/generate-docs.yaml
evalsocket: yes

evalsocket: Check it, In my action...i just uploaded 22MB cache https://github.com/evalsocket/flytectl/runs/2753321067?check_suite_focus=true#step:15:3

.github/workflows/generate-docs.yaml
evalsocket: that means we will copy these docs requirement and Make the command from the boilerplate. Right?

evalsocket: If yes then do it as a seprate pr

EngHabu (approved):

pmahindrakar-oss (approved):

PR #86, opened 2021-06-06
Added sandbox command

evalsocket:
# TL;DR
Introduce new command for making sandbox and serializing experience smooth

```bash
$ flytectl sandbox start # it will create sandbox and setup your flytectl/kubectl config. it also clone flytesnack repo
$ flytectl sandbox teardown # It will cleanup flyte sandbox env 
$ flytectl register examples/flytesnack # it will use flytesnacks release to read flyte_test_manifest and register all workflow available in the cluster. 
```

logs :
```
18:16:59 ➜ flytectl go run main.go sandbox start   
INFO[0000] Using config file: [config.yaml]             
🚀 It will take some time, We will start a fresh flyte cluster for you
{"status":"Pulling from flyteorg/flyte-sandbox","id":"dind"}
{"status":"Digest: sha256:05554e3ed90e2a009192f02150fce5cd007bb0f857712f11b789996929002752"}
{"status":"Status: Image is up to date for ghcr.io/flyteorg/flyte-sandbox:dind"}
Starting Docker daemon...
Done.
Starting k3s cluster...
Done.
Deploying Flyte...
...
...
...
-deployment.apps/flytepropeller condition met
👨‍💻 Flyte is ready! Flyte UI is available at http://localhost:30081/console. 🚀 🚀 🎉 
Please visit https://github.com/flyteorg/flytesnacks 🚀 
Register all flytesnacks example  'flytectl register examples  -d development  -p flytesnacks

```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


kumare3:
@evalsocket this is interesting. The only problem is, is this only docker based (though sandbox can be launched anywhere). But, I like this idea. One thing we need to be careful about is - docker client and how do we ensure we keep this updated etc (version compatibility and ensure that it works correctly)

pmahindrakar-oss:
Like the idea @evalsocket . Going away from user doing make and just doing all using flytectl would be good onboarding experience. We can simplify the steps further using this. Please mention the current onboarding and typical user flows and how they will change with this experience.

cmd/register/examples.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+
+	rconfig "github.com/flyteorg/flytectl/cmd/config/subcommand/register"
+	cmdCore "github.com/flyteorg/flytectl/cmd/core"
+	f "github.com/flyteorg/flytectl/pkg/filesystemutils"
+	"github.com/google/go-github/github"
+)
+
+const (
+	registerExampleShort = "Registers flytesnack example"
```
pmahindrakar-oss, line 18: nit : examples and same on L20

cmd/register/examples.go
```diff
@@ -0,0 +1,78 @@
...
+
+
+Usage
+`
+)
+
+func registerExampleFunc(ctx context.Context, args []string, cmdCtx cmdCore.CommandContext) error {
+	c := github.NewClient(nil)
+	opt := &github.ListOptions{Page: 1, PerPage: 1}
+	releases, _, err := c.Repositories.ListReleases(context.Background(), "flyteorg", "flytesnacks", opt)
```
pmahindrakar-oss, line 33: nit : org and repo constants

cmd/register/examples.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return err
+	}
+
+	jsonFile, err := os.Open(f.FilePathJoin(f.UserHomeDir(), ".flyte", "flytesnacks.yaml"))
+	if err != nil {
+		return err
+	}
+	defer jsonFile.Close()
+
```
pmahindrakar-oss, line 58: Whats the purpose of writing and reading the same file . I didn't find references to this yaml file anywhere else.
data from L43 can directly be used on L64

cmd/register/examples.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(byteValue, &FlyteSnacksRelease)
+	if err != nil {
+		return err
+	}
+	for _, v := range FlyteSnacksRelease {
+		rconfig.DefaultFilesConfig.Archive = true
```
pmahindrakar-oss, line 69: can be moved above the loop as it remains the same.

cmd/register/examples.go
```diff
@@ -0,0 +1,78 @@
...
+		args := []string{
+			fmt.Sprintf("https://github.com/flyteorg/flytesnacks/releases/download/%s/flytesnacks-%s.tgz", *releases[0].TagName, v.Name),
```
pmahindrakar-oss, line 71: Define the flytesnacks archive formatter constant

cmd/register/examples.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		if err := Register(ctx, args, cmdCtx); err != nil {
+			return err
```
pmahindrakar-oss, line 74: Wrap this error with addtional info about which registration failed based on the iteration value v

cmd/register/register_util.go
```diff
@@ -43,6 +43,21 @@ type HTTPClient interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
+var FlyteSnacksRelease []FlyteSnacks
+
+// FlyteSnacks Defines flyte test manifest structure
+type FlyteSnacks struct {
```
pmahindrakar-oss, line 49: nit : FlyteSnack and rename the var on L46 to be FlyteSnackReleases

cmd/sandbox/sandbox.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Long descriptions are whitespace sensitive when generating docs using sphinx.
+const (
+	sandboxShort = `Used for playing with sandbox.`
+	sandboxLong  = `
+Example Create sandbox cluster.
+::
+
+ flytectl sandbox cluster
```
pmahindrakar-oss, line 15: is this example correct, 

instead it should be 
`flytectl sandbox start`

`flytectl sandbox teardown`

cmd/sandbox/start.go
```diff
@@ -0,0 +1,139 @@
...
+type ExecResult struct {
+	StdOut   string
+	StdErr   string
+	ExitCode int
+}
+
+func startSandboxCluster(ctx context.Context, args []string, cmdCtx cmdCore.CommandContext) error {
+
+	fmt.Printf("%v It will take some time, Meanwhile enjoy your coffe %v \n", emoji.Rocket, emoji.Teapot)
+	response, err := http.Get("https://raw.githubusercontent.com/flyteorg/flytectl/master/config.yaml")
```
pmahindrakar-oss, line 49: nit : define the url as const

cmd/sandbox/start.go
```diff
@@ -0,0 +1,141 @@
...
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		if strings.Contains(scanner.Text(), "Flyte is ready! Flyte UI is available at http://localhost:30081/console") {
+			break
+		}
+		fmt.Println(scanner.Text())
+	}
+	fmt.Printf("%v Flyte is ready! Flyte UI is available at http://localhost:30081/console. %v %v %v \n", emoji.ManTechnologist, emoji.Rocket, emoji.Rocket, emoji.PartyPopper)
+	fmt.Printf("Please visit https://github.com/flyteorg/flytesnacks %v \n", emoji.Rocket)
+	return nil
```
pmahindrakar-oss, line 140: These lines seem would get executed even in failure case if flyte is not ready.

cmd/sandbox/teardown.go
```diff
@@ -0,0 +1,59 @@
...
+	f "github.com/flyteorg/flytectl/pkg/filesystemutils"
+)
+
+const (
+	teardownShort = "Teardown will cleanup the sandbox environment"
+	teardownLong  = `
+Teardown will remove docker container and all the flyte config 
+::
+
+ bin/flytectl teardown
```
pmahindrakar-oss, line 21: flytectl sandbox teardown

and add the other aliases.
flytectl sandbox stop

cmd/sandbox/teardown.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	for _, container := range containers {
+		if err := cli.ContainerStop(ctx, container.ID, nil); err != nil {
+			return err
+		}
+		if err := cli.ContainerRemove(ctx, container.ID, types.ContainerRemoveOptions{}); err != nil {
+			return err
+		}
+	}
```
pmahindrakar-oss, line 47: Are these calls idempotent.
ie. if the call the sandbox teardown fails for some reason, do these getting executed always result in container being stopped and removed if they arent already

cmd/sandbox/teardown.go
```diff
@@ -0,0 +1,59 @@
...
+
+	err = os.Remove(f.FilePathJoin(f.UserHomeDir(), ".flyte", "config.yaml"))
+	if err != nil {
+		return err
+	}
+	err = os.Remove(f.FilePathJoin(f.UserHomeDir(), ".flyte", "kube.yaml"))
+	if err != nil {
+		return err
+	}
```
pmahindrakar-oss, line 56: Move such config creations / destruction to separate file and reference them wherever needed.

cmd/register/examples.go
evalsocket: from this manifest, we will create a release asset URL. You can check flytesnack github release

cmd/sandbox/teardown.go
evalsocket: yes it's a bug i will fix it

cmd/sandbox/start.go
```diff
@@ -0,0 +1,130 @@
...
+	reader, err := cli.ContainerLogs(context.Background(), resp.ID, types.ContainerLogsOptions{
+		ShowStderr: true,
+		ShowStdout: true,
+		Timestamps: false,
+		Follow:     true,
+	})
+	if err != nil {
+		return err
+	}
+	scanner := bufio.NewScanner(reader)
```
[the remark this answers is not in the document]
evalsocket: do you have any suggestions for handling this ?

cmd/sandbox/start.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+	r, err := cli.ImagePull(ctx, ImageName, types.ImagePullOptions{})
+	if err != nil {
+		return err
+	}
+
+	if _, err := io.Copy(os.Stdout, r); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
pmahindrakar-oss: Can we use the ask prompt and instead of asking the user to run another command, run it internally after prompting him.

cmd/sandbox/start.go
```diff
@@ -0,0 +1,88 @@
...
+				fmt.Printf("%v Flyte is ready! Flyte UI is available at http://localhost:30081/console. %v %v %v \n", emoji.ManTechnologist, emoji.Rocket, emoji.Rocket, emoji.PartyPopper)
+				fmt.Printf("Please visit https://github.com/flyteorg/flytesnacks for more example %v \n", emoji.Rocket)
+				fmt.Printf("Register all flytesnacks example by running 'flytectl register examples  -d development  -p flytesnacks' \n")
+				break
+			}
+			fmt.Println(scanner.Text())
+		}
+		return nil
+	}
+	fmt.Println("Something goes wrong. We are not able to start sandbox container, Please check your docker client and try again \n", emoji.Rocket)
```
pmahindrakar-oss, line 86: Log the err

cmd/sandbox/start.go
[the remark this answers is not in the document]
pmahindrakar-oss: Can we get the container exist status which tell this instead of just relying on the log output which seems can contain both stderr and stdout .  

Otherwise , juts having some content in stderr should return a non-zero exit status

config.yaml
```diff
@@ -1,7 +1,8 @@
 admin:
   # For GRPC endpoints you might want to use dns:///flyte.myexample.com
-  endpoint: dns:///flyte.myexample.com
-  authType: Pkce
+  endpoint: dns:///localhost:30081
+  insecure: true
+#  authType: Pkce
```
pmahindrakar-oss, line 5: Exclude this file change

cmd/sandbox/start.go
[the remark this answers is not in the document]
evalsocket: done

config.yaml
evalsocket: why, I am planing to use same config file for sandbox

pmahindrakar-oss: Ok . can you remove the commented out line then.

cmd/sandbox/sandbox_util.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return resp.ID, nil
+}
+
+func watchError(cli *client.Client, id string) {
+	statusCh, errCh := cli.ContainerWait(context.Background(), id, container.WaitConditionNotRunning)
+	select {
+	case err := <-errCh:
+		if err != nil {
+			panic(err)
```
pmahindrakar-oss, line 133: Return this error instead of panic

pmahindrakar-oss (approved):

cmd/sandbox/sandbox_util.go
evalsocket: Rather than returning error from goroutine, I will handle panic in defer.

pmahindrakar-oss (approved):

kumare3:
This is awesome, I had one suggestion @EngHabu  / @evalsocket / @pmahindrakar-oss. what if we also published a version manifest with flytesnacks, that say, min platform version required?
That way we can ensure that the right version of platform is also deployed to sandbox?

pmahindrakar-oss:
Here the scenario which might be applicable to what @kumare3 wants.
User has deployed a sandbox already  V1 and playing around with it and there is new flytesnacks release which requires a V2 version of the platform

Now when he register the flytesnacks examples it would fail in some unknown manner since we allow the registration to go through without platform version check. I faced this issue when registering flytesnacks example which would run into circlular dependency error and on deploying the latest admin the error was fixed.

Inorder to solve this for the user , i agree we would need some sort of minimum version required with every flytesnacks release which can be tracked in manifest as released platform version

We can use the tags to define this mapping https://github.com/flyteorg/flyte/tags

The manifest file initially can contain just one attribute for now 
`min_platform_version=<flyte-tag-version>` from the tags link.
This change in flytesnacks should be verified in precommits by deploying the flyte-tag-version and registering all  flytesnacks examples.

In flytectl when registering the examples, it would check the version of the manifest and verify the deployed sandbox version.
The image https://github.com/orgs/flyteorg/packages/container/package/flyte-sandbox should additionally be tagged with release version instead of just the commit tag.

It would do semantic version comparison between the deployed sandbox image and the manifest version in flytesnacks and prompt the user to upgrade the cluster (i.e teardown and redeploy a new version)

@kumare3 let me know if this would work

kumare3:
This is right just one small change. Instead of the tag - @evalsocket added the version api in FlyteAdmin and flytectl. We could use the admin api to get the platform version

pmahindrakar-oss:
Thanks @kumare3 , but also we would want to solve this problem outside sandbox environment.

Can the serialised version of the files themselves contain some information on which version of the platform they should be registered in.  That way it can work across any environment.

kumare3:
That's interesting but flytekit does not know this today, but maybe we should add this info to serialized form and flytekit

pmahindrakar-oss (approved):

pmahindrakar-oss:
The version change would be taken as a followup.

PR #93, opened 2021-06-14
new flytestd libe version added

evalsocket:
# TL;DR
- New version of flytestdlib will remove extra login in flytectl

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


docs/source/index.rst
```diff
@@ -14,7 +14,7 @@ Install flytectl with homebrew tap
 
 .. prompt:: bash
 
-   brew tap flyteorg/homebrew-tap/flytectl
+   brew install flyteorg/homebrew-tap/flytectl
```
pmahindrakar-oss, line 17: The latest version of flytectl is 0.1.11 but when i run this command it tries to install an older version 0.1.6

```
brew install flyteorg/homebrew-tap/flytectl
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 4 taps (flyteorg/tap, hashicorp/tap, homebrew/core and homebrew/cask).
==> New Formulae
hashicorp/tap/consul-template                                                             hashicorp/tap/consul-terraform-sync
==> Updated Formulae
Updated 71 formulae.
==> Updated Casks
Updated 20 casks.
==> Deleted Casks
mapillary-uploader

==> Installing flytectl from flyteorg/tap
==> Downloading https://ghcr.io/v2/homebrew/core/go/manifests/1.16.5
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/go/blobs/sha256:416c5e2b7247c78482a5465f79d83c0240ee0c9098c8c7429f9c7af073402cc9
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:416c5e2b7247c78482a5465f79d83c0240ee0c9098c8c7429f9c7af073402cc9?se=2021-06-15T09%3A35%3A00Z&
######################################################################## 100.0%
==> Downloading https://github.com/flyteorg/flytectl/releases/download/v0.1.11/flytectl_Darwin_x86_64.tar.gz
==> Downloading from https://github-releases.githubusercontent.com/231630997/9d027e80-cd56-11eb-8e72-ad2677065b36?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AKIAIWNJYAX4CS
######################################################################## 100.0%
Error: flytectl 0.1.6 is already installed
```

.github/workflows/build.yaml
```diff
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: "2"
```
pmahindrakar-oss, line 16: If this change doesn't work to remove `Unchanged files with check annotations` then remove this and try to find another solution that can work for it

docs/source/index.rst
evalsocket: Please do `brew update` before running install

pmahindrakar-oss: Lets add that to the docs aswell for existing user who already have flyte installed.

```
brew upgrade flyteorg/tap/flytectl
```

evalsocket: i think we don't need that

.github/workflows/build.yaml
evalsocket: ok

docs/source/index.rst
pmahindrakar-oss: My point is more towards making the two options of installing flytectl in sync.

1] curl -s https://raw.githubusercontent.com/lyft/flytectl/master/install.sh | bash
Irrespective of when you execute this will always install the latest. It will overwrite the existing version that exists.(stateless installation)

2] brew install flyteorg/homebrew-tap/flytectl 
It will install what was brought by brew the last time it was executed. So this becomes stateful . 
Hence to make it sync with whats provided by the first command and also to eliminate any statefulness it would be good to have upgrade and install instructions available to the user in this second option. If there is way to combine the two into one command that would be good aswell. 

Please correct if you disagree.

evalsocket: Yes i agree...The first one is more portable and in docs, i added both option https://github.com/flyteorg/flytectl/blob/master/docs/source/index.rst#install

I think regarding brew, We don't need to add it in our docs about updates. Every project assume that you are aware about it

pmahindrakar-oss: Would be good for any of the naive users of brew to directly see the instructions in the doc. It doesn't harm. More the to the point instructions the  merrier it is :) as he won't have to find other ways to install/upgrade.
Power users definitely wont have a problem but i think we should have docs with naive users in mind .

Just my take on it. @EngHabu @kumare3 what you all think.

pmahindrakar-oss (approved):

docs/source/index.rst
evalsocket: done added those changes in readme

EngHabu (approved):

PR #105, opened 2021-06-17
Only pass literals to execution if they users provided values

EngHabu:
Signed-off-by: Haytham Abuelfutuh <haytham@afutuh.com>

# TL;DR
If users do not provide a value for an input, do not make a default empty value. This will enable admin to fill in default values if desired or error if the input is required instead of silently taking empty inputs as values.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/cmd/create/serialization_utils.go
+++ b/cmd/create/serialization_utils.go
@@ -1,35 +1,60 @@
 package create
 
 import (
+	"fmt"
+
 	"github.com/flyteorg/flyteidl/clients/go/coreutils"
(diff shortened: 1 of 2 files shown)


katrogan (dismissed):
are there any unit tests for these?

kumare3:
I think we should add unit tests for this

katrogan (approved):
oh boy so many types

PR #132, opened 2021-07-05
Added separate release workflow for releasing the binary

evalsocket:
# TL;DR
- Added separate release workflow for releasing the binary

Bump Version workflow will run on master merge and create a new tag
Test PR :- https://github.com/evalsocket/flytectl/pull/1 (bump version using title)
Result: https://github.com/flyteorg/flytectl/releases/tag/v0.2.0
Workflow :- https://github.com/evalsocket/flytectl/runs/2989584279?check_suite_focus=true

Release workflow will run in two cases
 - If bump version workflow create a tag
 - If We manually create a tag
Result : https://github.com/evalsocket/flytectl/releases/tag/v1.1.0
Workflow :- https://github.com/evalsocket/flytectl/runs/2989474510?check_suite_focus=true

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


.github/workflows/master.yml
```diff
@@ -22,42 +22,4 @@ jobs:
...
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: "0"
-      - name: Run GoReleaser dry run
-        uses: goreleaser/goreleaser-action@v2
-        with:
-          version: latest
-          args: release --rm-dist
-        env:
-          GITHUB_TOKEN: ${{ secrets.FLYTE_BOT_PAT }}
```
pmahindrakar-oss, line 42: Wont the removal of this now cause the automatic patch release to be stopped.
We should do the release similar to flyteidl where within the PR title itself.
```
- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. More details
```

pmahindrakar-oss:
We should support both manual release as well as PR based releases
Can you add details on how this change makes both of them work .
Also if you can test this in a cloned repo and provide the details on how it works in the PR.
* Looking for title based flytectl version bump .
* Without any version in title do patch releases
* and also manual releases .
Can you please add details for these along with simple test which shows this works

.github/workflows/master.yml
evalsocket: no both are independent bump version will only create a tag(patch release). The other release workflow triggers on tag event and run goreleaser.

In this case, release workflow will run for both case 
- If bump version workflow create a tag
- If We manually create a tag

pmahindrakar-oss: Clarified offline that the version tagged will be created for the minor version aswell in the title of the PR. Major release is currently manual.

And the release works if any version tag is present irrespective of manually created or through a github workflow.

pmahindrakar-oss (approved):

PR #135, opened 2021-07-06
Add pflag binding and fix loading of config

EngHabu:
Signed-off-by: Haytham Abuelfutuh <haytham@afutuh.com>

# TL;DR
Bind config pflags to be able to call flytectl and override configs through cmd.
e.g.
```prompt
flytectl --admin.endpoint dns://flyte.mycompany --insecure=false get projects
```

## Type
- [ ] Bug Fix
- [X] Feature
- [ ] Plugin

## Are all requirements met?

- [X] Code completed
- [X] Smoke tested
- [X] Unit tests added
- [X] Code documentation added
- [X] Any pending items have an associated Issue

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1226


cmd/root.go
```diff
@@ -47,6 +47,8 @@ func newRootCmd() *cobra.Command {
 
 	rootCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", "config file (default is $HOME/.flyte/config.yaml)")
 
+	configAccessor.InitializePflags(rootCmd.PersistentFlags())
```
pmahindrakar-oss, line 50: Can we exclude certain flags from being registered .
For eg : These might cause confusion with persistent flags of -p, -d and -o defined in flytectl

```
      --root.domain string                        Specified the domain to work on.
      --root.output string                         Specified the output type. (default "TABLE")
      --root.project string                        Specifies the project to work on.
```

pmahindrakar-oss (approved):
This is awesome. Thanks @EngHabu as this greatly helps the UX for flytectl.

cc : @evalsocket

cmd/root.go
EngHabu: we can remove the pflags generation for that... I've updated the pr

pmahindrakar-oss (approved):

PR #141, opened 2021-07-09
Graphviz does not like hyphens #patch

kumare3:
# TL;DR
Some graphs that have nested conditionals can have node names that use hyphens. This PR makes it so that hyphens in cluster name are also replaced with `_`

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added


wild-endeavor (approved):

EngHabu (approved):
Sweet!

PR #154, opened 2021-07-26
Render descriptions in flytectl

mayitbeegh:
# TL;DR
- Display input output variable description in the table view of `get task` `get workflow` `get launchplan`.
- Display input output variable description as comment when generating an execution spec template.

`get launchplan`
Screen Shot 2021-08-02 at 11 17 58

`get workflow`
Screen Shot 2021-08-02 at 11 18 18

`get task`
Screen Shot 2021-08-02 at 11 21 13

`get task --exexFile exec.yaml` before vs after
Screen Shot 2021-07-29 at 16 44 15

## Type
- [ ] Bug Fix
- [x] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1281

## Follow-up issue
_NA_

### DIFF
--- a/cmd/get/execution_util.go
+++ b/cmd/get/execution_util.go
@@ -3,6 +3,7 @@ package get
 import (
 	"errors"
 	"fmt"
+	"gopkg.in/yaml.v3"
 	"io/ioutil"
 	"os"
 
@@ -10,21 +11,19 @@ import (
 	"github.com/flyteorg/flyteidl/clients/go/coreutils"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
-
-	"sigs.k8s.io/yaml"
 )
 
-// ExecutionConfig is duplicated struct from create with the same structure. This is to avoid the circular dependency.
+// ExecutionConfig is duplicated struct from create with the same structure. This is to avoid the circular dependency. Only works with go-yaml.
 // TODO : replace this with a cleaner design
 type ExecutionConfig struct {
-	TargetDomain    string                 `json:"targetDomain"`
-	TargetProject   string                 `json:"targetProject"`
-	KubeServiceAcct string                 `json:"kubeServiceAcct"`
-	IamRoleARN      string                 `json:"iamRoleARN"`
-	Workflow        string                 `json:"workflow,omitempty"`
-	Task            string                 `json:"task,omitempty"`
-	Version         string                 `json:"version"`
-	Inputs          map[string]interface{} `json:"inputs"`
+	IamRoleARN      string               `yaml:"iamRoleARN"`
+	Inputs          map[string]yaml.Node `yaml:"inputs"`
+	KubeServiceAcct string               `yaml:"kubeServiceAcct"`
+	TargetDomain    string               `yaml:"targetDomain"`
+	TargetProject   string               `yaml:"targetProject"`
+	Task            string               `yaml:"task,omitempty"`
+	Version         string               `yaml:"version"`
+	Workflow        string               `yaml:"workflow,omitempty"`
 }
 
 func WriteExecConfigToFile(executionConfig ExecutionConfig, fileName string) error {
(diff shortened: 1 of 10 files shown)


pkg/printer/printer.go
```diff
@@ -164,6 +170,69 @@ func printJSONYaml(format OutputFormat, v interface{}) error {
...
+func FormatVariableDescriptions(variableMap map[string]*core.Variable) {
+	keys := make([]string, 0, len(variableMap))
+	// sort the keys for testing and consistency with other output formats
+	for k := range variableMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var descriptions []string
+	//for k, v := range variableMap {
```
katrogan, line 182: delete?

pkg/printer/printer.go
```diff
@@ -164,6 +170,69 @@ func printJSONYaml(format OutputFormat, v interface{}) error {
...
+			descriptions = append(descriptions, getTruncatedLine(fmt.Sprintf("%s: %s", k, v.Description)))
+		} else {
+			descriptions = append(descriptions, getTruncatedLine(k))
+		}
+
+	}
+	variableMap[DefaultFormattedDescriptionsKey] = &core.Variable{Description: strings.Join(descriptions, "\n")}
+}
+
+func FormatParameterDescriptions(variableMap map[string]*core.Parameter) {
```
katrogan, line 199:
```suggestion
func FormatParameterDescriptions(parameterMap map[string]*core.Parameter) {
```

pkg/printer/printer.go
```diff
@@ -164,6 +170,69 @@ func printJSONYaml(format OutputFormat, v interface{}) error {
...
+	// sort the keys for testing and consistency with other output formats
+	for k := range variableMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var descriptions []string
+	//for k, v := range variableMap {
+	for _, k := range keys {
+		if k == DefaultFormattedDescriptionsKey {
```
katrogan, line 184: why is this check necessary? shouldn't the variable map just be straight from the entity closure?

pkg/printer/printer.go
```diff
@@ -164,6 +170,69 @@ func printJSONYaml(format OutputFormat, v interface{}) error {
...
+			descriptions = append(descriptions, getTruncatedLine(fmt.Sprintf("%s: %s", k, v.Var.Description)))
+		} else {
+			descriptions = append(descriptions, getTruncatedLine(k))
+		}
+	}
+	variableMap[DefaultFormattedDescriptionsKey] = &core.Parameter{Var: &core.Variable{Description: strings.Join(descriptions, "\n")}}
+}
+
+func getTruncatedLine(line string) string {
+	// maybe add width to function signature later
```
katrogan, line 228:
```suggestion
	// TODO: maybe add width to function signature later
```

katrogan (approved):
lgtm! will let @pmahindrakar-oss review too

pkg/printer/printer.go
mayitbeegh: oops

pkg/printer/printer.go
mayitbeegh: There're some side effects in the tests that I don't fully understand. For example, if I change the order of the tests, or if I try to run tests individually, things break. This is one of the problems I ran into, the function under test got called twice.

cmd/get/execution_util.go
```diff
@@ -6,25 +6,25 @@ import (
...
 	"os"
 
+	"gopkg.in/yaml.v3"
+
 	cmdUtil "github.com/flyteorg/flytectl/pkg/commandutils"
 	"github.com/flyteorg/flyteidl/clients/go/coreutils"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
-
-	"sigs.k8s.io/yaml"
```
pmahindrakar-oss, line 14: For consistency we should probably convert other cli structs to also use the new gopkg.
Can you create an issue for it .

pkg/printer/printer.go
pmahindrakar-oss: @mayitbeegh If any test execution changes based on the order then that would be case if you are using global state across all the tests.
In such case move common state changes in the setup function
and override behavior for test specific states after the setup is called in the test. That should fix such cases.
We have done this across most of the tests in flytectl.
Also is it possible for you to just write unit tests covering this function, the bot is showing that this function is not covered anywhere

pmahindrakar-oss (approved):

pkg/printer/printer.go
mayitbeegh: @pmahindrakar-oss Okay, I can fix that

pmahindrakar-oss (approved):

PR #157, opened 2021-08-06
Remove input output columns for list workflows

mayitbeegh:
# TL;DR
The workflow interface is not returned for listing requests, which makes the get workflow table look rather empty sometimes:
Screen Shot 2021-08-05 at 19 25 42

This PR removes the input output columns from the table if list workflow is called while keeping the table intact for when fetching a single workflow.
Screen Shot 2021-08-06 at 14 30 35

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/cmd/get/workflow.go
+++ b/cmd/get/workflow.go
@@ -92,6 +92,12 @@ var workflowColumns = []printer.Column{
 	{Header: "Created At", JSONPath: "$.closure.createdAt"},
 }
 
+var listWorkflowColumns = []printer.Column{
+	{Header: "Version", JSONPath: "$.id.version"},
+	{Header: "Name", JSONPath: "$.id.name"},
+	{Header: "Created At", JSONPath: "$.closure.createdAt"},
+}
+
 func WorkflowToProtoMessages(l []*admin.Workflow) []proto.Message {
 	messages := make([]proto.Message, 0, len(l))
 	for _, m := range l {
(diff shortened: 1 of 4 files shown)


cmd/get/workflow.go
```diff
@@ -146,32 +157,34 @@ func getWorkflowFunc(ctx context.Context, args []string, cmdCtx cmdCore.CommandC
...
 	}
-	return adminPrinter.Print(config.GetConfig().MustOutputFormat(), workflowColumns, WorkflowToProtoMessages(workflows)...)
+	return adminPrinter.Print(config.GetConfig().MustOutputFormat(), listWorkflowColumns, WorkflowToProtoMessages(workflows)...)
 }
 
 // FetchWorkflowForName fetches the workflow give it name.
 func FetchWorkflowForName(ctx context.Context, fetcher ext.AdminFetcherExtInterface, name, project,
-	domain string) ([]*admin.Workflow, error) {
+	domain string) ([]*admin.Workflow, bool, error) {
 	var workflows []*admin.Workflow
```
katrogan, line 168: nit: we can use named return variables here, which will initialize them with their zero values so we don't have to do so manually: https://yourbasic.org/golang/named-return-values-parameters/

katrogan (approved):
looks good! will let @pmahindrakar-oss give the final approval 

(also why is code cov bot so confused wrt unchanged lines? 😩)

mayitbeegh:
> looks good! will let @pmahindrakar-oss give the final approval
> 
> (also why is code cov bot so confused wrt unchanged lines? 😩)

🤷

cmd/get/workflow.go
mayitbeegh: Fixed

pmahindrakar-oss:
Hi @katrogan @mayitbeegh , there is an issue we are aware of in flytectl bot and @evalsocket had a fix for this but it didn't work the last time he had tried it.

Adding him to the thread here to check if he has got a chance to look into it

pmahindrakar-oss (approved):

evalsocket:
@pmahindrakar-oss @mayitbeegh @katrogan  Currently not sure how to remove them from github diff because they added it as [feature](https://github.blog/2018-12-14-introducing-check-runs-and-annotations/)

but it is possible to remove them from [codcov]( https://docs.codecov.com/docs/github-checks (Possible to disable Unchanged files with check annotations in codecov using codecov yaml config))

pmahindrakar-oss:
Would be good to keep the line by line coverage feature because that one we don't want tot turn off for the changed files

But for the unchanged files the beta check for such annotations need to be disabled .Can we check if there is explicit way to just do this

Here's another reference I found 

https://github.com/reviewdog/reviewdog/issues/932

PR #161, opened 2021-08-19
Bug fix in sandbix start 

evalsocket:
# TL;DR
- Updated boilerplate 
-  Small bug fix in version compare until (Currently we have a condition that checks greater then equal but now we only check GreateThen.....This bug exist in version compare)
- Because we are mounting container file in localhost system that's why we are not able to change the permission of the file, But now we created a empty file and we only overwrite while mounting docker volume

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [ ] Code completed
- [ ] Smoke tested
- [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_
OR

### DIFF
--- a/cmd/sandbox/start.go
+++ b/cmd/sandbox/start.go
@@ -47,7 +47,7 @@ Mount your source code repository inside sandbox
 
  bin/flytectl sandbox start --source=$HOME/flyteorg/flytesnacks 
 	
-Run specific version of flyte, Only available after v0.14.0+
+Run specific version of flyte, Only available after v0.13.0+
 ::
 
  bin/flytectl sandbox start  --version=v0.14.0
--- a/cmd/sandbox/start.go
+++ b/cmd/sandbox/start.go
@@ -55,7 +55,7 @@ Run specific version of flyte, Only available after v0.14.0+
 Usage
 	`
 	k8sEndpoint                  = "https://127.0.0.1:30086"
-	flyteMinimumVersionSupported = "v0.14.0"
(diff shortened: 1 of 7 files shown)


pkg/util/util.go
```diff
@@ -48,7 +52,7 @@ func IsVersionGreaterThan(version1, version2 string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return semanticVersion2.LessThanOrEqual(semanticVersion1), nil
+	return semanticVersion1.GreaterThan(semanticVersion2), nil
```
pmahindrakar-oss, line 55: Lets  add unit test with the bug symptoms  and before the change that test should fail and after change it should pass.

cmd/sandbox/start.go
```diff
@@ -55,7 +55,7 @@ Run specific version of flyte, Only available after v0.14.0+
 Usage
 	`
 	k8sEndpoint                  = "https://127.0.0.1:30086"
-	flyteMinimumVersionSupported = "v0.14.0"
+	flyteMinimumVersionSupported = "v0.13.0"
```
pmahindrakar-oss, line 58: Lets update the docs also to mention this if not already.

pkg/util/util.go
evalsocket: ok

pmahindrakar-oss (approved):

PR #163, opened 2021-08-19
add --dryRun flag to flytectl subcommands

hamersaw:
# TL;DR
Adds --dryRun flag to flytectl subcommands which enables execution without making any modifications.

## Type
- [ ] Bug Fix
- [X] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [X] Smoke tested
- [ ] Unit tests added

## Complete description
Adds the --dryRun flag to all 'create', 'delete', 'register', and 'update' subcommands in flytectl. This option allows execution of all operations, encompassing both client and server side, without making any modifications. Any operation which requires change of any kind is skipped.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/901

## Follow-up issue
_NA_


cmd/create/execution_util.go
```diff
@@ -10,16 +10,24 @@ import (
...
 
 func createExecutionRequestForWorkflow(ctx context.Context, workflowName, project, domain string,
 	cmdCtx cmdCore.CommandContext) (*admin.ExecutionCreateRequest, error) {
 	// Fetch the launch plan
-	lp, err := cmdCtx.AdminFetcherExt().FetchLPVersion(ctx, workflowName, executionConfig.Version, project, domain)
-	if err != nil {
-		return nil, err
+	var lp *admin.LaunchPlan
+	var err error
+	if executionConfig.DryRun {
```
EngHabu, line 23: This might stir some controversy :-D... I think `--dryRun` is ok to make server side calls as long as it doesn't have any side-effects...
e.g. [Kubectl docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run)

> Dry run mode helps to evaluate a request through the typical request stages (admission chain, validation, merge conflicts) up until persisting objects to storage.

In this case, I would say retrieving something from Admin is ok to do... as long as we don't create an actual execution...

EngHabu (commented):
Awesome! Just a minor comment

cmd/create/execution_util.go
hamersaw: Sure, I was thinking this would pave the way for a future dryRun=client / dryRun=server split. Now that I think a bit, it is probably a better idea to get as much coverage as possible. I added all the server-side communication that is possible without modifications.

kumare3:
ohh - DryRun that talks to server :). Do we need to update all the api's to support dryrun or like a header?

EngHabu (approved):

PR #164, opened 2021-08-20
Update upgrade.go

mouuff:
## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [ ] Smoke tested
- [ ] Unit tests added

## Complete description
We don't need to call Rollback() after calling Update() (unless we want to rollback after a successful update of course)

### DIFF
--- a/cmd/upgrade/upgrade.go
+++ b/cmd/upgrade/upgrade.go
@@ -102,7 +102,7 @@ func upgrade(u *updater.Updater) (string, error) {
 		}
 		return fmt.Sprintf("Successfully updated to version %s", latestVersion), nil
 	}
-	return "", u.Rollback()
+	return "", nil
 }
 
 func isUpgradeSupported(goos platformutil.Platform) (bool, error) {


EngHabu (dismissed):

EngHabu:
Please fix DCO

cmd/upgrade/upgrade.go
```diff
@@ -102,7 +102,7 @@ func upgrade(u *updater.Updater) (string, error) {
 		}
 		return fmt.Sprintf("Successfully updated to version %s", latestVersion), nil
 	}
-	return "", u.Rollback()
+	return "", nil
```
evalsocket, line 105: Why do we need this change ? 
```
# if this condition failed then it is failed upgrade and we want to roll back.
if updateStatus == updater.Updated {
}
```

mouuff: I have no idea why this comment is there, but this is not the case.
If the update fails then go-rocket-update will call `Rollback()` itself if needed.
And if the update fails then it will return an error... here we call rollback when the call to `Update()` returns no error and when status is not `Updated`.
This means if the status is `UpToDate` you are calling `Rollback()` (why?)

evalsocket: got it thanks

evalsocket (approved):

PR #200, opened 2021-10-15
Workflow: Do not bump version and consequently release on docs or boilerplate update

radiantly:
Instead of checking whether the committer is the flyte-bot, we add the docs and boilerplate directories to the ignored paths list. Thus, if all changed files reside in the ignored directories, the version is not bumped and consequently the release workflow is not run.

### Type

N/A

### Are all requirements met?

N/A

### Tracking Issue
Fixes https://github.com/flyteorg/flyte/issues/1621

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -4,18 +4,20 @@ on:
   push:
     branches:
       - master
+    paths-ignore:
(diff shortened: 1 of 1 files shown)


samhita-alla:
cc: @evalsocket

PR #212, opened 2021-11-07
[Housekeeping] Refactor flytectl flags

aeioulisa:
# TL;DR
Flytectl code has one issue. Internal method/pkg use flags directly rather than using them as parameters. We should only use flags in the root command. After that, we should pass the value in the method.

## Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [X] Code completed
- [X] Smoke tested
- [ ] Unit tests added

## Complete description
Just refactor.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1617

## Follow-up issue
_NA_

### DIFF
--- a/cmd/register/register_util.go
+++ b/cmd/register/register_util.go
@@ -23,7 +23,6 @@ import (
 	"github.com/google/go-github/github"
 
 	"github.com/flyteorg/flytectl/cmd/config"
-	rconfig "github.com/flyteorg/flytectl/cmd/config/subcommand/register"
 	cmdCore "github.com/flyteorg/flytectl/cmd/core"
 	"github.com/flyteorg/flytectl/pkg/printer"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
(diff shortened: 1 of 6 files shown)


pmahindrakar-oss:
The description and the changes don't seems to be on the same lines.

Can you clarify since I just see variable and import path renames

cmd/register/register_util.go
```diff
@@ -215,15 +214,15 @@ func hydrateIdentifier(identifier *core.Identifier) {
...
+		identifier.Version = defaultFilesConfig.Version
 	}
 }
 
 func hydrateTaskSpec(task *admin.TaskSpec, sourceCode string) error {
 	if task.Template.GetContainer() != nil {
 		for k := range task.Template.GetContainer().Args {
 			if task.Template.GetContainer().Args[k] == "" || task.Template.GetContainer().Args[k] == registrationRemotePackagePattern {
-				remotePath, err := getRemoteStoragePath(context.Background(), Client, rconfig.DefaultFilesConfig.SourceUploadPath, sourceCode, rconfig.DefaultFilesConfig.Version)
+				remotePath, err := getRemoteStoragePath(context.Background(), Client, defaultFilesConfig.SourceUploadPath, sourceCode, defaultFilesConfig.Version)
```
pmahindrakar-oss, line 225: I think the main agenda was to not use such config directly in these internal functions and be used only the first level command function .
Rest of the places we want to pass the config as parameter wherever it's used

aeioulisa: @pmahindrakar-oss thanks for your review
I've been pushed a new commit
Could you please review it?

aeioulisa:
@pmahindrakar-oss thanks for your review
I've been pushed a new commit
Could you please review it?

pmahindrakar-oss (approved):
Looks good to me. Thanks for making those changes .

evalsocket (approved):

PR #224, opened 2021-11-30
Changes in cmd folder

SmritiSatyanV:
Changes to files in 'cmd' folder of flytectl- grammatical error and restructuring.


cmd/register/examples.go
```diff
@@ -15,18 +16,18 @@ import (
 const (
 	registerExampleShort = "Registers flytesnack example"
 	registerExampleLong  = `
-Registers all latest flytesnacks example
+The below command registers the latest flytesnacks example:
```
pmahindrakar-oss, line 19: I would prefer to keep the earlier sentence here and in other places . @samhita-alla  what do you think
I see the same pattern changes which seems more verbose.

samhita-alla: Since this content gets displayed in the docs, I think what Smriti has written would be more instructive. What do you say, @pmahindrakar-oss?

pmahindrakar-oss: As long as it conveys what the command does , the more  concise the better .
eg linux man page for ls https://man7.org/linux/man-pages/man1/ls.1.html

Would prefer a command docs which is similar to such man pages. I feel having this verbiage `The below command` and similar things in other places is making it too verbose. We can probably correct the ones which are ambiguous or not clear

samhita-alla: Right. @SmritiSatyanV, can we have concise descriptions? Please make sure the concise descriptions talk about the command clearly. For example for flytectl register:

```
Register all examples using FlyteCTL. This command is useful to make the examples accessible in the Flyte UI. It uploads the packages protobufs to the Flyte backend.
... 
```

Does this look okay, @pmahindrakar-oss?

pmahindrakar-oss: Sure looks good to me. Thanks @samhita-alla

SmritiSatyanV: Sure

SmritiSatyanV:
Creating new PR

PR #241, opened 2021-12-18
Improve flytectl config init UX 

evalsocket:
# TL;DR
- In `config init`, Added logic for timing host name (Now host will trim `dns://`, `https://` & `http://` )
- Now in `config init` default value of insecure would be false

Testing 
- flytectl config init --host=google (Will fail)
- flytectl config init  --host=127.0.0.1 (will pass)
- flytectl config init  --host=localhost (will pass)
- flytectl config init  --host=127.0.0.1.0 (will fail)

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1880

## Follow-up issue
_NA_
OR

### DIFF
--- a/cmd/config/subcommand/config/init_flags.go
+++ b/cmd/config/subcommand/config/init_flags.go
@@ -3,7 +3,7 @@ package config
 //go:generate pflags Config --default-var DefaultConfig --bind-default-var
 var (
 	DefaultConfig = &Config{
-		Insecure: true,
+		Insecure: false,
 		Storage:  false,
 	}
 )
--- a/cmd/configuration/configuration.go
+++ b/cmd/configuration/configuration.go
@@ -51,6 +51,8 @@ var prompt = promptui.Select{
 	Items: []string{"S3", "GCS"},
 }
 
+var dnsPrefix = [3]string{"dns://", "http://", "https://"}
+
 // CreateConfigCommand will return configuration command
 func CreateConfigCommand() *cobra.Command {
 	configCmd := viper.GetConfigCommand()
@@ -82,7 +84,7 @@ func initFlytectlConfig(ctx context.Context, reader io.Reader) error {
 	templateStr := configutil.GetSandboxTemplate()
 
 	if len(initConfig.DefaultConfig.Host) > 0 {
-		templateValues.Host = fmt.Sprintf("dns:///%v", initConfig.DefaultConfig.Host)
+		templateValues.Host = fmt.Sprintf("dns://%s", trim(initConfig.DefaultConfig.Host))
 		templateStr = configutil.AdminConfigTemplate
 		if initConfig.DefaultConfig.Storage {
 			templateStr = configutil.GetAWSCloudTemplate()
@@ -113,4 +115,12 @@ func initFlytectlConfig(ctx context.Context, reader io.Reader) error {
 	}
 	fmt.Printf("Init flytectl config file at [%s]", configutil.ConfigFile)
 	return nil
+}
+
+func trim(hostname string) string {
+	for _, prefix := range dnsPrefix {
+		hostname = strings.Trim(hostname, prefix)
+	}
+	return hostname
+
 }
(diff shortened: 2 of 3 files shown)


cmd/configuration/configuration.go
```diff
@@ -51,6 +51,8 @@ var prompt = promptui.Select{
 	Items: []string{"S3", "GCS"},
 }
 
+var dnsPrefix = [3]string{"dns://", "http://", "https://"}
```
pmahindrakar-oss, line 54: dnsPrefix -> endpointPrefixes

cmd/config/subcommand/config/init_flags.go
```diff
@@ -3,7 +3,7 @@ package config
 //go:generate pflags Config --default-var DefaultConfig --bind-default-var
 var (
 	DefaultConfig = &Config{
-		Insecure: true,
+		Insecure: false,
```
pmahindrakar-oss, line 6: Can we also document this in flytectl docs where ever these options are referenced

cmd/configuration/configuration.go
```diff
@@ -114,3 +117,11 @@ func initFlytectlConfig(ctx context.Context, reader io.Reader) error {
...
 	return nil
 }
+
+func trim(hostname string) string {
+	for _, prefix := range dnsPrefix {
+		hostname = strings.Trim(hostname, prefix)
+	}
+	return hostname
+
+}
```
pmahindrakar-oss, line 127: Lets add unit tests for this function

cmd/configuration/configuration.go
```diff
@@ -77,12 +79,13 @@ func initFlytectlConfig(ctx context.Context, reader io.Reader) error {
...
 	templateValues := configutil.ConfigTemplateSpec{
 		Host:     "dns:///localhost:30081",
-		Insecure: initConfig.DefaultConfig.Insecure,
+		Insecure: true,
 	}
 	templateStr := configutil.GetSandboxTemplate()
 
 	if len(initConfig.DefaultConfig.Host) > 0 {
-		templateValues.Host = fmt.Sprintf("dns:///%v", initConfig.DefaultConfig.Host)
+		templateValues.Host = fmt.Sprintf("dns://%s", trim(initConfig.DefaultConfig.Host))
```
pmahindrakar-oss, line 87: After setting this can we validate that its valid dns endpoint

cmd/config/subcommand/config/init_flags.go
pmahindrakar-oss: Also does this cause any changes to the sandbox config.

evalsocket: no

cmd/configuration/configuration.go
evalsocket: we don't need that extra check

pmahindrakar-oss (dismissed):

cmd/configuration/configuration.go
pmahindrakar-oss: It would be useful if user enters garbage data eg : https://0.0.0.0.0.0
Url.Parser might be able to catch it i suppose.

cmd/configuration/configuration_test.go
```diff
@@ -56,8 +56,22 @@ func TestSetupConfigFunc(t *testing.T) {
 	assert.Nil(t, initFlytectlConfig(ctx, yes))
 	assert.Nil(t, initFlytectlConfig(ctx, yes))
 	assert.Nil(t, initFlytectlConfig(ctx, no))
-	initConfig.DefaultConfig.Host = "test"
+	initConfig.DefaultConfig.Host = "flyte.org"
```
pmahindrakar-oss, line 59: Can we also add a test with localhost:30081 ie one with host:port

pmahindrakar-oss (approved):
We shouldn't allow updates like these too. 
```
admin:
  endpoint: dns://localhost:30081:1
```
Not a high priority though. We should see if there is already a dns endpoint validation library in apimachinery

PR #244, opened 2021-12-20
Generate random version when registering workflows

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
Generate a random version if users don't specify a version

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [ ] Code completed
- [ ] Smoke tested
- [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1968

## Follow-up issue
_NA_


evalsocket (commented):
Please add the documentation for version, Currently it's a required parameter in register

cmd/config/subcommand/register/files_config.go
```diff
@@ -11,7 +11,7 @@ var (
 
 // FilesConfig containing flags used for registration
 type FilesConfig struct {
-	Version              string `json:"version" pflag:",version of the entity to be registered with flyte which are un-versioned after serialization."`
+	Version              string `json:"version" pflag:",version of the entity to be registered with flyte which are un-versioned after serialization. Random UUID if empty"`
```
evalsocket, line 14: version of the entity to be registered with flyte which are un-versioned after serialization, By default it generate a random uid for version.

pingsutw: updated, thanks

EngHabu:
Hey @pingsutw i don't think random versioning is useful. I understand it works around the annoyance of local iterations but I think it breaks the premise of immutability and idempotency.
I think an important characteristic in any versioning scheme is determinism. One thing you can use is either hashing of the individual resource you're about to register or hashing of the entire archive file.. either way would be ok...

pingsutw:
Make sense, will use the hashing of resource spec

PR #280, opened 2022-02-16
Fixed validation logic for launch plan schedule registration

pmahindrakar-oss:

# TL;DR
Launchplan schedule registration needs a validation so as to  not face runtime issues with missing parameters.
Hence during the registration the check ensures all the parameters have corresponding default or fixed values.

The defaultInput params in flyte are divided into two types  - Required or one that have default values.
Any required params need to have a Fixed value for a scheduled launchplan.
Earlier it was expecting both the types to have a fixed value which was incorrect

This PR corrects this validation logic to differentiate the required and default ones.

Also improves the UX for the error message to display which params are missing the values.

```
 --------------------------------------------------------------------------------------------------------------------- --------- -------------------------------------------------------- 
| /Users/praful/flytesnacks/cookbook/core/_pb_output/152_my_cron_scheduled_lp_3.pb                                    | Failed  | Error hydrating spec due to param values are missing   |
|                                                                                                                     |         | on scheduled workflow for the following params [name]. |
|                                                                                                                     |         | Either specify them having a default or fixed value    |
 --------------------------------------------------------------------------------------------------------------------- --------- -------------------------------------------------------- 
```

## Type
- [X] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [X] Code completed
- [X] Smoke tested
- [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2129
## Follow-up issue
_NA_


cmd/register/register_util.go
```diff
@@ -275,34 +275,76 @@ func hydrateTaskSpec(task *admin.TaskSpec, sourceCode, sourceUploadPath, version
 	return nil
 }
 
-func validateLaunchSpec(lpSpec *admin.LaunchPlanSpec) error {
-	if lpSpec == nil {
+func validateLPWithSchedule(lpSpec *admin.LaunchPlanSpec) error {
+	if lpSpec.EntityMetadata == nil || lpSpec.EntityMetadata.Schedule == nil {
```
katrogan, line 279: do you still want to keep the check for if the spec itself is nil before dereferencing?

cmd/register/register_util.go
```diff
@@ -275,34 +275,76 @@ func hydrateTaskSpec(task *admin.TaskSpec, sourceCode, sourceUploadPath, version
...
-		var scheduleFixedParams []string
-		if lpSpec.DefaultInputs != nil {
-			for paramKey := range lpSpec.DefaultInputs.Parameters {
-				if paramKey != schedule.KickoffTimeInputArg {
-					scheduleFixedParams = append(scheduleFixedParams, paramKey)
-				}
+	schedule := lpSpec.EntityMetadata.Schedule
+	var scheduleRequiredParams []string
+	if lpSpec.DefaultInputs != nil {
+		for paramKey, paramValue := range lpSpec.DefaultInputs.Parameters {
```
katrogan, line 285: I don't think that this is right, default inputs are okay - they already have a value associated with them that can be used at execution time.

Ideally we should check the workflow spec and compare any unbound inputs (outside of KickoffTimeInputArg) and then raise that delta as the error.

cmd/register/register_util.go
katrogan: ah nvm I see it's been moved below.

cmd/register/register_util.go
pmahindrakar-oss: Updated @katrogan . Please review when you get a chance

pmahindrakar-oss: @katrogan I found this confusing.
FixedInputs *core.LiteralMap 
We have lpSpec.DefaultInputs which are basically *core.ParameterMap  and each Parameter can be 
// Types that are valid to be assigned to Behavior:
	//	*Parameter_Default
	//	*Parameter_Required
So we have lpSpec.DefaultInputs which can be of Parameter_Default or Parameter_Required.
Only if its Parameter_Default , does it have Literal value which makes sense because thats the default value for the parameter.

But if its a Parameter_Required then its doesn't have a literal value .
Can we not have required params with default values ?
The name DefaultInputs confuses me . Can you help clarify on this aspect

katrogan: we can't have required parameters with default values. they are required exactly because there is no bound value 

In the launch plan spec, anything in the DefaultInputs message can be overridden in the create execution request, or otherwise the values in the DefaultInputs block will be used. I *think* that's why we called it DefaultInputs because we default to using those values. Maybe the naming could use a refresh before flyteidl 1.0.0 :)

PR #318, opened 2022-04-27
flyetcl version should ignore talking to admin if host is not configures

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
- Ignore talking to the admin if the host is not configured
- Do not prompt users to install new flytectl if the current version is the same as the latest version
```
 A new release of flytectl is available: v0.5.21 → v0.5.21 

{
  "App": "flytectl",
  "Build": "fc7b1dd",
  "Version": "v0.5.21",
  "BuildTime": "2022-04-27 16:25:14.783715 +0800 CST m=+0.018003092"

```

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [ ] Code completed
- [ ] Smoke tested
- [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2256

## Follow-up issue
_NA_

### DIFF
--- a/cmd/core/cmd.go
+++ b/cmd/core/cmd.go
@@ -61,6 +61,10 @@ func generateCommandFunc(cmdEntry CommandEntry) func(cmd *cobra.Command, args []
 		}
 
 		adminCfg := admin.GetConfig(ctx)
+		if len(adminCfg.Endpoint.String()) == 0 {
+			return cmdEntry.CmdFunc(ctx, args, CommandContext{})
+		}
+
 		clientSet, err := admin.ClientSetBuilder().WithConfig(admin.GetConfig(ctx)).
 			WithTokenCache(pkce.TokenCacheKeyringProvider{
 				ServiceUser: fmt.Sprintf("%s:%s", adminCfg.Endpoint.String(), pkce.KeyRingServiceUser),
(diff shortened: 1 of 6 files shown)


pmahindrakar-oss:
This doesn't solve the issue here 
https://github.com/flyteorg/flyte/issues/2256

As it will still print those auth error messages .

What we need is to avoid talking to admin completely if the host is not configured in all paths while running version command.

You might want to take a look at the command context creation where this is being done and this is coming from imported IDL code and hence we might want to fix this issue there

pingsutw:
@pmahindrakar-oss Thanks, I just updated it, so now we will not initialize admin clients if the host is None.

cmd/core/cmd.go
```diff
@@ -61,6 +61,10 @@ func generateCommandFunc(cmdEntry CommandEntry) func(cmd *cobra.Command, args []
 		}
 
 		adminCfg := admin.GetConfig(ctx)
+		if len(adminCfg.Endpoint.String()) == 0 {
+			return cmdEntry.CmdFunc(ctx, args, CommandContext{})
```
kumare3, line 65: I do not think this is the solution right. What we want is that client should be lazy to connect and ignore if needed

pingsutw: Do you mean that some of the commands (like sandbox, upgrade) should not connect to FlyteAdmin no matter if `adminCfg.Endpoint` is set or not?

pmahindrakar-oss: @kumare3 we can have current solution implemented and then take up the lazy initialization as an enhancement. @pingsutw you can create a separate issue for lazy initialization for the client.
This can be interim solution

pmahindrakar-oss (approved):

PR #331, opened 2022-06-20
Add commands to fetch active and archived launchplans

SmritiSatyanV:
Issue: https://github.com/flyteorg/flyte/issues/2620
Slack conversation: https://flyte-org.slack.com/archives/CP2HDHKE1/p1655433353538459
Signed-off-by: SmritiSatyanV <smriti@union.ai>


docs/source/gen/flytectl_compile.rst
```diff
@@ -0,0 +1,104 @@
+.. _flytectl_compile:
```
pmahindrakar-oss, line 1: Can you check if this file is added in some index .

cmd/get/launch_plan.go
```diff
@@ -55,6 +55,16 @@ Retrieve all the launch plans with filters:
 
   flytectl get launchplan -p flytesnacks -d development --filter.fieldSelector="name=core.basic.lp.go_greet"
 
+Retrieve all active launch plans with filters:
```
pmahindrakar-oss, line 58: nit : you can omit the `with filters` in both the lines.

We probably want to add all properties which are filterable in the docs not just for launchplan, but for executions, workflows, task , named entity etc and this should be available easily on docs.flyte.org

Can we create a separate ticket to capture this effort

pmahindrakar-oss (approved):

PR #355, opened 2022-09-23
Specify cluster pool when creating execution

iaroslav-ciupin:
Signed-off-by: Iaroslav Ciupin <iaroslav@union.ai>

## Read then delete

- Use #patch, #minor #majora or #none in the pull-request title to bump the corresponding version. Otherwise, the patch version
  will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)

# TL;DR
When creating an execution via flytectl, user should be able to pass custom cluster pool to which execution should be assigned.

## Type
- [x] Feature

## Are all requirements met?

- [x] Code completed
- [ ] Smoke tested
- [x] Unit tests added

## Complete description


cmd/create/execution.go
```diff
@@ -195,16 +200,24 @@ func createExecutionCommand(ctx context.Context, args []string, cmdCtx cmdCore.C
...
-		if executionRequest, err = createExecutionRequestForWorkflow(ctx, execParams.name, sourceProject, sourceDomain, cmdCtx, executionConfig, targetExecName); err != nil {
+		executionRequest, err = createExecutionRequestForWorkflow(ctx, execParams.name, sourceProject, sourceDomain, cmdCtx, executionConfig, targetExecName)
+		if err != nil {
 			return err
 		}
 	default:
 		return fmt.Errorf("invalid execution type %v", execParams.execType)
 	}
+
+	if executionRequest.Spec.ClusterAssignment == nil {
```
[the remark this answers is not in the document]
iaroslav-ciupin: spec is always initialized in https://github.com/flyteorg/flytectl/blob/master/cmd/create/execution_util.go#L155
If its nil, it will fail on server anyway, because what's the point of sending ExecutionRequest without spec?

cmd/create/execution.go
```diff
@@ -132,6 +133,12 @@ The generated file would look similar to this. Here, empty values have been dump
   task: core.type_system.custom_objects.add
   version: v3
 
+9. When creating a new execution, you can assign it to a specific cluster pool:
+
+::
+
+   flytectl create execution --execFile execution_spec.yaml -p flytesnacks -d development --clusterPool my-gpu-cluster
```
pmahindrakar-oss, line 140: Can we mention that this requires implementing the executor plugin and link to plugin interface

cmd/create/execution.go
```diff
@@ -195,16 +200,22 @@ func createExecutionCommand(ctx context.Context, args []string, cmdCtx cmdCore.C
...
+		if err != nil {
 			return err
 		}
 	default:
 		return fmt.Errorf("invalid execution type %v", execParams.execType)
 	}
+
+	if executionConfig.ClusterPool != "" {
+		executionRequest.Spec.ClusterAssignment = &admin.ClusterAssignment{ClusterPoolName: executionConfig.ClusterPool}
+	}
```
pmahindrakar-oss, line 218: Can we contain this change to createExecutionRequest. Modifying the create objected in places outside it initial creation becomes confusing

pmahindrakar-oss (dismissed):

pmahindrakar-oss (approved):

PR #369, opened 2022-11-05
Add extra host to start command

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
[slack thread](https://flyte-org.slack.com/archives/CP2HDHKE1/p1667231330440659)
Related to https://github.com/spotify/docker-client/issues/995
Failed to connect to webhook in the demo on ubuntu because k8s can't resolve the host name `host.docker.internal`.

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [ ] Unit tests added
- [ ] Code documentation added
- [ ] Any pending items have an associated Issue

## Complete description
^^^
## Tracking Issue
NA
## Follow-up issue
_NA_

PR #380, opened 2023-01-11
Change extra hosts

wild-endeavor:
# TL;DR
I think this used to work because the flyte binary used to run as a process inside the top level container.  Now it runs in the cluster.  In order for K8s to call out the the webhook service (which in the `--dev` case is running on the user's host machine), we need to pipe through the host machine's IP.

This works in conjunction with https://github.com/flyteorg/flyte/pull/3228.

This has been tested with both the `--dev` switch and without it.

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [ ] Unit tests added

## Complete description
The extra hosts was first added in https://github.com/flyteorg/flytectl/pull/369.  At the time the flyte binary ran as a process in the dind container.  I guess at that time, K8s was able to see the dind container as 127.0.0.1, and so routing there would hit the flyte binary.

### DIFF
--- a/pkg/docker/docker_util.go
+++ b/pkg/docker/docker_util.go
@@ -50,7 +50,7 @@ var (
 	}
 	StdWriterPrefixLen = 8
 	StartingBufLen     = 32*1024 + StdWriterPrefixLen + 1
-	ExtraHosts         = []string{"host.docker.internal:127.0.0.1"}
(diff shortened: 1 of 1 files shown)


jeevb (approved):
🙌

pingsutw (approved):

PR #408, opened 2023-06-05
Support envs when creating execution

honnix:
# TL;DR
Support envs when creating execution.

## Type
- [ ] Bug Fix
- [x] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added

## Complete description
As described by https://github.com/flyteorg/flyte/issues/3746, `flytectl` currently does not support passing envs to flyteadmin when creating executions. This PR adds the support by utilizing execution spec file.

Note that this works differently than `pyflyte`, which is on one hand not ideal, but OTOH given the existence of execution spec file in `flytectl`, using the file instead of a plain json string is less error prone and more traceable.

## Tracking Issue
Closes https://github.com/flyteorg/flyte/issues/3746

## Follow-up issue
_NA_


eapolinario (approved):
Thank you. This looks great.

PR #411, opened 2023-07-03
Fix the error message for the update cmd

oliverhu:
Fix Flyte's #3806 [Housekeeping] Inconsistent treatment of project id param in flytectl

# TL;DR
As project name is required rather than project id, check name instead.

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [ ] Smoke tested
- [x] Unit tests added

## Tracking Issue
[https://github.com/flyteorg/flyte/issues/<number>](https://github.com/flyteorg/flyte/issues/3806)

### Updated behavior
```
❯ ./bin/flytectl update project -p flytesnack --id flytesnack3
Error: both project and id are passed
❯ ./bin/flytectl update project -p flytesnacks --name flytesnack3
Project flytesnacks updated
❯ ./bin/flytectl update project --id flytesnacks --name flytesnack3
Project flytesnacks updated```

### DIFF
--- a/cmd/config/subcommand/project/project_config.go
+++ b/cmd/config/subcommand/project/project_config.go
@@ -78,7 +78,7 @@ func (c *ConfigProject) GetProjectSpec() (*admin.Project, error) {
 
 	// Get projectId from file, if not provided, fall back to project
 	if len(projectSpec.Id) == 0 {
-		projectSpec.Id = config.GetConfig().Project
+		projectSpec.Id = id
 	}
 	return &projectSpec, nil
 }
--- a/cmd/config/subcommand/project/project_config.go
+++ b/cmd/config/subcommand/project/project_config.go
@@ -74,12 +74,12 @@ func (c *ConfigProject) GetProjectSpec(cf *config.Config) (*admin.Project, error
 
 	project := cf.Project
 	if len(projectSpec.Id) == 0 && len(project) == 0 {
-		err := fmt.Errorf(clierrors.ErrProjectNameNotPassed)
+		err := fmt.Errorf(clierrors.ErrProjectNotPassed)
 		return nil, err
 	}
 
 	if len(projectSpec.Id) > 0 && len(project) > 0 {
-		err := fmt.Errorf(clierrors.ErrProjectIdBothPassed)
+		err := fmt.Errorf(clierrors.ErrProjectIDBothPassed)
 		return nil, err
 	}
 
--- a/cmd/config/subcommand/project/project_config.go
+++ b/cmd/config/subcommand/project/project_config.go
@@ -48,7 +48,7 @@ var DefaultProjectConfig = &ConfigProject{
 // GetProjectSpec return project spec from a file/flags
 func (c *ConfigProject) GetProjectSpec() (*admin.Project, error) {
 	projectSpec := admin.Project{}
-	id := config.GetConfig().Project
+
 	if len(c.File) > 0 {
 		yamlFile, err := ioutil.ReadFile(c.File)
 		if err != nil {
@@ -71,6 +71,8 @@ func (c *ConfigProject) GetProjectSpec() (*admin.Project, error) {
 		}
 		projectSpec.State = projectState
 	}
+
+	id := config.GetConfig().Project
 	if len(projectSpec.Id) == 0 && len(id) == 0 {
 		err := fmt.Errorf(clierrors.ErrProjectNameNotPassed)
 		return nil, err
--- a/cmd/config/subcommand/project/project_config.go
+++ b/cmd/config/subcommand/project/project_config.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 
 	"github.com/flyteorg/flytectl/clierrors"
+	"github.com/flyteorg/flytectl/cmd/config"
 	"github.com/flyteorg/flytectl/pkg/filters"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: cmd/update/project.go [hunk 1]
```diff
@@ -88,8 +88,9 @@ func updateProjectsFunc(ctx context.Context, args []string, cmdCtx cmdCore.Comma
 	if err != nil {
 		return err
 	}
-	if projectSpec.Id == "" {
-		return fmt.Errorf(clierrors.ErrProjectNotPassed)
+
+	if projectSpec.Name == "" {
```
katrogan, line 92: I think this isn't quite correct, since we do need the id to be specified to uniquely identify the project to update. However, we might not necessary want to update the name (this is just the human readable string) since the UpdateProject call does only updates modified fields on the backend.

I think what we should do is check that if the projectSpec.Id is empty then read it from the global flytectl `-p` project flag and if that's omitted, raise an error.

oliverhu: Ah, you're right. I thought `-p` is passed to `projectSpec.name`. Updating

oliverhu:
@katrogan  updated, I think `id` is not expected to be updated, so added a check for that. Output:

```
❯ ./bin/flytectl update project -p flytesnack --id flytesnack3
Error: Project `id` can't be updated. Hint: did you mean `-p` instead of `--id` to specify the project?
❯ ./bin/flytectl update project -p flytesnack --name flytesnack3
Project flytesnack failed to update due to rpc error: code = NotFound desc = project [flytesnack] not found
Error: rpc error: code = NotFound desc = project [flytesnack] not found
❯ ./bin/flytectl update project -p flytesnacks --name flytesnack3
Project flytesnacks updated
```

katrogan:
I think the issue is we also overwrite the project id to the global project flag: https://github.com/flyteorg/flytectl/blob/master/cmd/config/subcommand/project/project_config.go#L65 passed here: https://github.com/flyteorg/flytectl/blob/master/cmd/update/project.go#L87

Maybe we should only fallback to the global project flag when `--id` isn't specified

oliverhu:
Yeah, I saw that. It is very confusing.. what about we change:
`func (c *ConfigProject) GetProjectSpec(id string) (*admin.Project, error) { ` to `func (c *ConfigProject) GetProjectSpec() (*admin.Project, error) {`

So ProjectSpec will be a 1:1 mapping with the input.

Then we always get Project from `config.GetConfig().Project`?

cmd/config/subcommand/project/project_config.go
```diff
@@ -45,8 +46,9 @@ var DefaultProjectConfig = &ConfigProject{
 }
 
 // GetProjectSpec return project spec from a file/flags
-func (c *ConfigProject) GetProjectSpec(id string) (*admin.Project, error) {
+func (c *ConfigProject) GetProjectSpec() (*admin.Project, error) {
 	projectSpec := admin.Project{}
+	id := config.GetConfig().Project
```
katrogan, line 51: nit: can you define this closest to where it's used?

cmd/config/subcommand/project/project_config.go
```diff
@@ -56,23 +58,28 @@ func (c *ConfigProject) GetProjectSpec(id string) (*admin.Project, error) {
...
-	if err != nil {
+	if len(projectSpec.Id) == 0 && len(id) == 0 {
+		err := fmt.Errorf(clierrors.ErrProjectNameNotPassed)
 		return nil, err
 	}
-	projectSpec.State = projectState
+
+	// Get projectId from file, if not provided, fall back to project
+	if len(projectSpec.Id) == 0 {
+		projectSpec.Id = config.GetConfig().Project
```
katrogan, line 81:
```suggestion
		projectSpec.Id = id
```

cmd/update/project.go
```diff
@@ -84,10 +82,16 @@ Usage
...
 
 func updateProjectsFunc(ctx context.Context, args []string, cmdCtx cmdCore.CommandContext) error {
-	projectSpec, err := project.DefaultProjectConfig.GetProjectSpec(config.GetConfig().Project)
+	projectSpec, err := project.DefaultProjectConfig.GetProjectSpec()
+
 	if err != nil {
 		return err
 	}
+
+	if project.DefaultProjectConfig.ID != "" {
```
katrogan, line 91: I think we no longer need to check if this is empty since `GetProjectSpec` should now check and error

cmd/config/subcommand/project/project_config.go
katrogan: Also I think we should use one consistent way to set the project id flag, rather than using the global project flag here but the id flag in else block below, can we exclusively read the `ConfigProject` id field instead? And either error  if the global project flag is set too (or fall back to it when the `ConfigProject` ID isn't set)

katrogan (commented):
Thank you for this!

cmd/config/subcommand/project/project_config.go
oliverhu: I think it already behaves as you described, it only uses global project flag if ConfigProject ID isn't set

katrogan (dismissed):

katrogan:
mind fixing up the lint & unit test failures?

katrogan (approved):

PR #417, opened 2023-07-30
Fix missing Type & State info for get launchplans cmd

pbrogan12:
# TL;DR

The `get launchplan` command references non-existent fields in the `LaunchPlan` message when generating output, resulting in no data for the `TYPE` and `STATE` columns.

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [ ] Unit tests added

## Complete description

Referenced the correct fields.

## Follow-up issue
_NA_


kumare3 (commented):
The type field here seems odd, the original expectation seems to be task type which is weird and now adding launchplan type is not very useful 😞

PR #418, opened 2023-08-21
fix(flytectl): Update FetchWorkflowLatestVersion to fetch the latest version of a workflow

LeHyperion:
# TL;DR

Fix the `FetchWorkflowLatestVersion` function in flytectl to return the latest workflow version.

## Type

 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Complete description

The `FetchWorkflowLatestVersion` function in `workflow_fetcher.go` has been updated to retrieve the latest workflow version by using the last element in the `wVersions` slice. This change ensures that the most recent version is fetched when calling this function.

## Tracking Issue

fixes [3972](https://github.com/flyteorg/flyte/issues/3972)

### DIFF
--- a/pkg/ext/workflow_fetcher.go
+++ b/pkg/ext/workflow_fetcher.go
@@ -49,7 +49,7 @@ func (a *AdminFetcherExtClient) FetchWorkflowLatestVersion(ctx context.Context, 
 	if err != nil {
 		return nil, err
 	}
-	return a.FetchWorkflowVersion(ctx, name, wVersions[0].Id.Version, project, domain)
+	return a.FetchWorkflowVersion(ctx, name, wVersions[len(wVersions)-1].Id.Version, project, domain)
 }
 
 // FetchWorkflowVersion fetches particular version of workflow


LeHyperion:
> would you be willing to make a few more changes? I think this function is wrong, or the name is wrong. we should loop through the filters, and if there isn't a limit filter we should add a limit of 1 to the existing list of filters, and then we should sort https://docs.flyte.org/en/latest/concepts/admin.html#sorting-syntax. Is this something you could add please?

I'm new to contributing, so I appreciate your patience with my inquiries.

1. Does flytectl have a built-in filter function?
2. How can I obtain the sorting key?
3. For retrieving the newest workflow version, is the sequence FetchWorkflowLatestVersion -> Sort by Key-> Filter Limit 1 correct?
4. Does fetching the latest launch plan follow a similar process?

wild-endeavor:
Oh no worries, thank you very much @LeHyperion!

1. Does flytectl have a built-in filter function?  so filters are just [strings](https://github.com/flyteorg/flyteidl/blob/7f62a157921db31d5f7a2285b4e9a42b4a01eaa4/protos/flyteidl/admin/common.proto#L104) in the service definition, so [admin will expect](https://github.com/flyteorg/flyteadmin/blob/6c97bb9c5558af240b2d67290bb35393990678cb/pkg/manager/impl/execution_manager.go#L1432) strings to be there but it looks like flytectl does implement its own[ filter object](https://github.com/flyteorg/flytectl/blob/145bc5c01db75a21254626ecba0e01d6dbea7851/pkg/filters/type.go#L13) which is what's used in the [FetchWorkflowLatestVersion](https://github.com/flyteorg/flytectl/blob/145bc5c01db75a21254626ecba0e01d6dbea7851/pkg/ext/workflow_fetcher.go#L46) function.

Looking at this more, I think what's happening is that [this filter object](https://github.com/flyteorg/flytectl/blob/145bc5c01db75a21254626ecba0e01d6dbea7851/cmd/get/workflow.go#L188) is not correct.  If we just change it, it should work.

3. How can I obtain the sorting key?  I don't think you need to obtain the sort key.  Looking at the api, you should be able to specify the sort in the `Filter` object.

4. For retrieving the newest workflow version, is the sequence FetchWorkflowLatestVersion -> Sort by Key-> Filter Limit 1 correct? hmm not sure what you mean.  but see the answer to 1.  I think that's all that's needed.  maybe we make another Filter object to represent latest.

5. Does fetching the latest launch plan follow a similar process? Yup - could you make it there too?

Thanks again.

davidmirror-ops:
@LeHyperion just checking in to see if you need any help with this?

PR #436, opened 2023-10-20
Use proto file name to infer message type

honnix:
# TL;DR
Use proto file name to infer message type to have more precise unmarshalling.

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added (the existing unit tests cover this change well)

## Complete description
flytekit generated proto file name has a pattern to precisely indicate the message type: `.*_1.pb`  is task, `.*_2.pb` is workflow and `.*_3.pb` is launch plan. We can rely on this pattern to do more precise unmarshalling. Note that we still try other types if the unmarshalling based on the precisely matched type does not work.

This PR keeps backward compatibility so if the pattern is not seen,  it goes through all possible types and tries one by one.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3042

## Follow-up issue
_NA_

### DIFF
--- a/cmd/register/register_util.go
+++ b/cmd/register/register_util.go
@@ -92,6 +92,8 @@ var projectColumns = []printer.Column{
 	{Header: "Additional Info", JSONPath: "$.Info"},
 }
 
+// Regex to match file name like xxx_1.pb, xxx_2.pb, or xxx_3.pb, and the subgroup catches the number 1, 2 or 3
+// This is used to match proto files created by pyflyte, where xxx_1.pb is a task spec, xxx_2.pb is a workflow spec, and xxx_3.pb is launch plan
 var fnameRegex = regexp.MustCompile(`^.*_(?P<index>[1-3])\.pb$`)
 
 type unMarshalFunc = func(ctx context.Context, fileContents []byte, fname string, errCollection errors2.ErrorCollection) (proto.Message, error)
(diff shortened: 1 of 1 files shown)


cmd/register/register_util.go
```diff
@@ -92,37 +92,72 @@ var projectColumns = []printer.Column{
...
+func reorderUnMarshalFuncs(fname string) []unMarshalFunc {
+	if match := fnameRegex.FindStringSubmatch(fname); match != nil {
+		indexStr := match[fnameRegex.SubexpIndex("index")]
+		index, err := strconv.Atoi(indexStr)
+		if err != nil {
+			panic(fmt.Sprintf("unexpected error when coverting [%s] to int, file name [%s]", indexStr, fname))
+		}
+
+		var reordered []unMarshalFunc
+		for i, f := range unMarshalFuncs {
```
honnix, line 150: This loop basically bubbles up the matched function to the first place.

cmd/register/register_util.go
```diff
@@ -92,37 +92,72 @@ var projectColumns = []printer.Column{
 	{Header: "Additional Info", JSONPath: "$.Info"},
 }
 
+var fnameRegex = regexp.MustCompile(`^.*_(?P<index>[1-3])\.pb$`)
```
pingsutw, line 95: nit: we could add some comments here

honnix: Added in a latter commit.

honnix:
Thank you for approving. Could you please help merge it when ready? Thanks.

PR #438, opened 2023-10-28
Fix project status change on update for archived projects

mcanueste:
# TL;DR
Currently, updating some field (i.e. `Description`) of an `Archived` project causes the project state to
change to `Active`. This is caused by a bug in the `updateProjectsFunc` function used for the `update` command.

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added

## Complete description
During project updates, `edited`, a default project created based on the given configs/flags,
comes with `admin.Project_ACTIVE` state by default if both `activate` and `archive` flags 
have not been set. 

Then a `copy` of the target project will be updated based on `edited` and then used for showing
the diff. In the current implementation, the state of the `copy` is set to be the same as `edited` 
projects state, which is `Project_ACTIVE` if no flags are set.

Also, after showing the diff, the `edited` is used for updating the project. Since it comes
with `Project_ACTIVE` state by default, the target project will be updated to have `Project_ACTIVE`
if both `activate` and `archive` flags are unset, regardless of the projects previous state.

On this PR, we make sure we set the correct state to both `copy` and `edited` by checking 
if  both `activate` and `archive` flags are set/unset. If one of the flags are set, we update target
project state (`copy`) based on the `edited` state since it will have the desired state. If both
flags are unset, we set the `edited` state to be the same as target (`copy`) state. This way, updating
`archived` projects will not change their state.

This is a bit hacky IMO. Proper solution would be to refactor `project.ConfigProject` and this file 
in order to separate the logic of setting `ConfigProject` struct fields from creation of a 'default' 
project based on those flags. Having a proper order of precedence between global config, 
YAML file input, and the flags for `ConfigProject` would also	be good. Since the changes
would be too big in that case, I have not worked on this kind of solution. However, if needed
I can create follow-up issues and work on them. 

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4114

## Follow-up issue
_NA_

### DIFF
--- a/cmd/update/project.go
+++ b/cmd/update/project.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/admin"
+
 	"github.com/flyteorg/flytectl/clierrors"
 	"github.com/flyteorg/flytectl/cmd/config"
 	"github.com/flyteorg/flytectl/cmd/config/subcommand/project"
@@ -85,7 +86,9 @@ Usage
 )
 
 func updateProjectsFunc(ctx context.Context, args []string, cmdCtx cmdCore.CommandContext) error {
-	edits, err := project.DefaultProjectConfig.GetProjectSpec(config.GetConfig())
+	projectConfig := project.DefaultProjectConfig
+
+	edits, err := projectConfig.GetProjectSpec(config.GetConfig())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


samhita-alla:
@eapolinario, could you please review this PR?

wild-endeavor (approved):
thanks!

wild-endeavor:
oh can you lint please?

mcanueste:
@wild-endeavor sorry for the delay.

I fixed the linting error on the `cmd/update/project.go` file since it was the file I edited, also rebased the branch to the latest master commit to make PR a bit easier.

There is also a linting error on `cmd/get/launch_plan.go` file when I run `make lint` locally. The error message is like this:
```bash
cmd/get/launch_plan.go:132:115: string `.description` has 5 occurrences, make it a constant (goconst)
        {Header: "Outputs", JSONPath: "$.closure.expectedOutputs.variables." + printer.DefaultFormattedDescriptionsKey + ".description"},
``` 

This PR doesn't touch `launch_plan.go` at all, and I don't know whether if I should attempt to fix that error as well since it is
out of scope for this PR.

wild-endeavor:
Can you give me write perms to your fork?  I will take a look.

wild-endeavor (approved):

PR #461, opened 2024-02-13
fix: Use ExtraHosts only for linux machines with native docker engine

MortalHappiness:
# TL;DR
Use ExtraHosts only for Linux machines with native docker engine

## Type
- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added

## Complete description
Detect whether `/var/run/docker.sock` is not a symbolic link and is a socket file. If true, add ExtraHosts. Otherwise, return empty ExtraHosts.

## Tracking Issue
None

## Follow-up issue
_NA_


pkg/docker/docker_util.go
```diff
@@ -60,6 +62,18 @@ var (
 	ExtraHosts         = []string{"host.docker.internal:host-gateway"}
 )
 
+// GetExtraHosts will return extra hosts for the container
+func GetExtraHosts() []string {
+	// Linux machine with native docker engine does not have host.docker.internal and needs to add the hostname mapping manually.
+	if runtime.GOOS == "linux" {
+		// Here assume that if /var/run/docker.sock exists and is a socket file and is not symbolic link, then the docker engine is native
+		if fileInfo, err := os.Lstat("/var/run/docker.sock"); err == nil && fileInfo.Mode().Type() == fs.ModeSocket {
+			return []string{"host.docker.internal:host-gateway"}
```
pingsutw, line 71:
```suggestion
			return ExtraHosts
```

MortalHappiness: Fixed

PR #466, opened 2024-03-22
Add ENV VAR FLTYE_ADMIN_ENDPOINT (#4948)

zychen5186:
## Read then delete

- Make sure to use a concise title for the pull-request.
- Use #patch, #minor #majora or #none in the pull-request title to bump the corresponding version. Otherwise, the patch version
  will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [ ] Code completed
- [ ] Smoke tested
- [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_
OR


cmd/core/cmd.go
```diff
@@ -66,6 +69,18 @@ func generateCommandFunc(cmdEntry CommandEntry) func(cmd *cobra.Command, args []
 		}
 
 		adminCfg := admin.GetConfig(ctx)
+
+		if len(os.Getenv("FLTYE_ADMIN_ENDPOINT")) > 0 {
+			envEndpoint, err := url.Parse(os.Getenv("FLTYE_ADMIN_ENDPOINT"))
+			if err != nil {
+				fmt.Println("Error parsing URL:", err)
```
troychiu, line 76: I think we can use fmt.Errorf here

cmd/config/env_var_reader.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/flyteorg/flyte/flyteidl/clients/go/admin"
+	"github.com/flyteorg/flyte/flytestdlib/config"
+)
+
+const flyteAdminEndpoint = "FLYTE_ADMIN_ENDPOINT"
+
+type FuncType func() error
+
+var funcMap = map[string]FuncType{flyteAdminEndpoint: updateAdminEndpoint}
```
troychiu, line 17: Could we have a name which represent the meaning instead of funcMap?

cmd/config/env_var_reader.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+
+	"github.com/flyteorg/flyte/flyteidl/clients/go/admin"
+	"github.com/flyteorg/flyte/flytestdlib/config"
+)
+
+const flyteAdminEndpoint = "FLYTE_ADMIN_ENDPOINT"
```
troychiu, line 13: @pingsutw do we typically follow the pattern of ending with "EnvVar"? Ex: flyteAdminEndpointEnvVar

cmd/config/env_var_reader.go
```diff
@@ -0,0 +1,45 @@
...
+
+type FuncType func() error
```
troychiu, line 15: Same for the name

cmd/config/env_var_reader.go
```diff
@@ -0,0 +1,45 @@
...
+
+var funcMap = map[string]FuncType{flyteAdminEndpoint: updateAdminEndpoint}
+
+func UpdateConfigWithEnvVar() error {
+	for envVar, f := range funcMap {
```
troychiu, line 20: f -> sth like configUpdateFunc

cmd/config/env_var_reader.go
```diff
@@ -0,0 +1,45 @@
...
+			if err := f(); err != nil {
+				return fmt.Errorf("error update config with env var: %v", err)
+			}
+		}
+	}
+	return nil
+}
+
+func updateAdminEndpoint() error {
+	ctx := context.Background()
```
troychiu, line 31: If we have multiple env var in the future, will initializing ctx every single time be a problem?

cmd/config/env_var_reader.go
zychen5186: How do you think about envToUpdater

cmd/config/env_var_reader.go
zychen5186: FuncType -> Updater, based on the proposed map name

EngHabu:
All fields that show up in cli are also settable through env vars.
If it's --connection.endpoint them it'll be FLYTE_CONNECTION_ENDPOINT or something like that... You shouldn't need this PR IMO.

zychen5186:
Hi @EngHabu, thanks for pointing this out! I also saw this `v.viper.AutomaticEnv()` in flytestdlib that seems to bind flags with env var, since the corresponding flag I was trying to assign is --admin.endpoint, I tried both FLYTE_ADMIN_ENDPOINT and ADMIN_ENDPOINT but didn't work. 
I found this  `err := v.bindViperConfigsFromEnv(r)` function in flytestdlib which I'm not so sure, but I think it might bind --admin.endpoint with ADMIN.ENDPOINT, which is not able to be set as an env var.

PR #473, opened 2024-04-23
Integrate Bubbletea Pagination into get execution

zychen5186:
# TL;DR
Use Bubbletea Pagination to show the get execution list instead of printing 100 rows each time

## Type
- [ ] Bug Fix
- [x] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [ ] Unit tests added

## Complete description
User can use -i or --interactive to trigger Bubbletea interface, for example,
```bash
make compile
./bin/flytectl get execution -p my-project -d development -i
```
 please refer to [Bubbletea Github page](https://github.com/charmbracelet/bubbletea/blob/master/README.md) to see what Bubbletea library offers.
Users can use the Bubbletea interface to check the entire list at once by going left and right, instead of showing 100 rows every command and need to specify which page they want to look at in the command line. 
image

## Tracking Issue
[https://github.com/flyteorg/flyte/issues/4440](https://github.com/flyteorg/flyte/issues/4440)

## Follow-up issue
Integrate this functionality into other flytectl get functions.


pkg/bubbletea/bubbletea_pagination_util.go
```diff
@@ -0,0 +1,174 @@
...
+	for _, m := range curShowMessage {
+		printableMessages = append(printableMessages, &PrintableProto{Message: m})
+	}
+
+	jsonRows, err := json.Marshal(printableMessages)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal proto messages")
+	}
+
+	var rawRows []interface{}
```
troychiu, line 95: Is this reusable?
https://github.com/flyteorg/flytectl/blob/64853531cb35f2a6214ff1cc82e9d39fc6109c08/pkg/printer/printer.go#L115

pkg/bubbletea/bubbletea_pagination_util.go
```diff
@@ -0,0 +1,174 @@
...
+	batchLen[batchIndex] = len(msg)
+
+	return msg
+}
+
+func Paginator(_listHeader []printer.Column, _callback DataCallback) {
+	listHeader = _listHeader
+	callback = _callback
+
+	msg := []proto.Message{}
```
troychiu, line 144:
```suggestion
	var msg []proto.Message
```

troychiu (commented):
This is awesome!

pkg/printer/printer.go
```diff
@@ -141,7 +142,9 @@ func (p Printer) JSONToTable(jsonRows []byte, columns []Column) error {
 	if r := printer.Render(headers, rows, positions, true); r == -1 {
 		return fmt.Errorf("failed to render table")
 	}
-	fmt.Printf("%d rows\n", len(rows))
+	if w == os.Stdout {
+		fmt.Printf("%d rows\n", len(rows))
+	}
```
zychen5186, line 147: This line is to avoid printing "%d rows\n" when using pagination

troychiu:
@katrogan @pmahindrakar-oss could you help review?

kumare3 (approved):
Thank you for following the pattern

pingsutw (approved):

pmahindrakar-oss (dismissed):

pmahindrakar-oss:
Can we restrict the flag to be used with only list api's and not for modification api's /  on usage display that this mode is not supported for the command being used

cmd/get/execution.go
```diff
@@ -146,6 +159,12 @@ func getExecutionFunc(ctx context.Context, args []string, cmdCtx cmdCore.Command
 		return err
 	}
 	logger.Infof(ctx, "Retrieved %v executions", len(executionList.Executions))
+
+	if config.GetConfig().Interactive {
+		bubbletea.Paginator(executionColumns, getCallBack(ctx, cmdCtx))
+		return nil
+	}
```
zychen5186, line 166: > Can we restrict the flag to be used with only list api's and not for modification api's / on usage display that this mode is not supported for the command being used

Currently, I'm planning to add pagination to only list API functions, and they will check if -i flag is specified, other non-list API functions will not be affected even if the -i flag is set. Do you think we should explicitly inform users that interactive CLI will not be triggered if they use -i on non-list commands?

troychiu: Since we will keep adding interactive feature, I think it's fine if the flag is not yet implemented for some apis at this moment. But we should explicitly point out what apis support interactive mode when this is done.

pmahindrakar-oss: i Agree since we will keep implementing new changes for this mode. We can probably mention the help string that this is only support for get calls and others is no-op

pmahindrakar-oss (approved):

